PR #92, opened 2019-07-08
Apply fix for "option_group: null"

mellogarrett:


types/atlas.ts
```diff
@@ -2,5 +2,5 @@ export interface AtlasQuestionOption {
   question_id: string;
   option_key: string;
   option_value: string;
-  option_group?: string;
+  option_group?: string | null;
```
MarkDKoslow: Hmm I'm wondering if this is cleaner or whether to change the way we translate options to choices to omit this field if it's null

mellogarrett: Yea I was thinking about that as well. let me go ahead and make the change there.

PR #98, opened 2019-07-19
Autocomplete dropdown

mellogarrett:
Refactored the Dropdown component to an Autocomplete component using `downshift`


components/dropdown-question/index.tsx
```diff
@@ -12,44 +17,106 @@ interface DropdownQuestionProps {
...
+                          selected: highlightedIndex === i,
+                        })}
+                      >
+                        {display || value}
+                      </MenuItem>
+                    );
+                  })}
+              </MenuList>
+            </Paper>
+          ) : null}
```
MarkDKoslow: Not a blocker - I think it'd improve the UX if we showed a message when the user's search term doesn't match any results. Something like "No results for 'xyz'. Please update your search query." Right now it looks like it shows a small horizontal bar 
Screen Shot 2019-07-22 at 1 26 28 PM

mellogarrett:
@MarkDKoslow I went ahead and implemented a "Your search returned no matches" element as well as some other small updates if you could kindly re-review. Thanks!

components/dropdown-question/index.test.tsx
```diff
@@ -140,4 +143,21 @@ describe('DropdownQuestionComponent', () => {
     fireEvent.change(getByLabelText(testQuestion.title), { target: { value: 'abc' } });
     expect(queryAllByRole('option').length).toBe(1);
   });
+
+  test('should show disabled option when search yields no results', () => {
```
MarkDKoslow: Nice! 👏

MarkDKoslow:
Looks great! Thanks for adding that 👏

PR #101, opened 2019-07-22
Fix & Enable Posting to Atlas for Sprint Retro Form

mellogarrett:
The `api.forms_submit_sprint_retrospective` Postgres function is defined with many positional arguments. Some are required, some are optional. If you want to skip an optional argument, for example, you must provide `null` for that argument position. Otherwise, you are basically trying to call a function that doesn't exist. Our frontend implementation was omitting `null` values from ever being sent to Atlas.


http/submit-form/backend-adapters/atlas/index.ts
```diff
@@ -25,8 +25,7 @@ const formatAtlasRequest = (form: Form, mergeId: string) => {
     const externalKey = question.externalKey;
     const answer = getAnswerForQuestion(question);
 
-    // Only include item in post request if question has an externalKey and value is non null
-    if (externalKey && answer) {
```
MarkDKoslow: Ah this would explain why it was working previously but didn't when I tried it more recently - I must have filled out all the nullable fields originally

PR #102, opened 2019-07-22
Add hotjar tracking code to all pages in forms app

MarkDKoslow:
New hotjar poll can be found here - https://insights.hotjar.com/sites/1155688/polls/list

It looks like we're already maxing out our page view allocation on our current plan. I'm not sure whether this will affect whether the poll is shown to all users but I'll check with Josh. It looks like we get 10K page views / day on our current plan, so forms alone could account for a large percentage of that allocation on Fridays.


MarkDKoslow:
Talked to Josh and he confirmed that the poll probably won't be shown to all users because we're maxing out our page allocation. Right now it'd be show to 37.9% of students

mellogarrett:
Good to know! @boblauer are you okay with that?

PR #166, opened 2019-08-22
Fuzzy search

mellogarrett:
- Use [fuse.js](https://github.com/krisk/Fuse) for fuzzy search
- Fix `<Linkify>` issues with React keys

_unrelated to title, but big change that was blocking me from creating this PR_
- Replace Husky pre-commit hooks with `"test": "npm-run-all test:*"` which can be seen in [other projects](https://github.com/LambdaSchoolEngineering/atlas-db/blob/17514eee8da1ffbd4be82e033cc94ee677cab934/package.json#L46-L49). Husky kept complaining that it couldn't find my version of Node and wouldn't let me commit. Seems to be a bug with Husky & nvm integration. No idea why it's just now coming up... Anyway, as long as we run `npm run test` before we push commits, everything will be fine and dandy. If we don't, Travis will yell at us.


boblauer:
Looks like Travis threw up, not sure what's going on there.

components/dropdown-question/index.tsx
```diff
@@ -42,10 +45,7 @@ export const DropdownQuestionComponent = ({ question, onChange, style, className
         selectedItem,
         clearSelection,
       }) => {
-        const suggestions = choices.filter((choice) => {
-          const lowercaseDisplayValue = (choice.display || choice.value).toString().toLowerCase();
-          return inputValue === null || lowercaseDisplayValue.includes(inputValue.toLowerCase());
-        });
+        const suggestions = inputValue ? fuse.search(inputValue) : choices;
```
MarkDKoslow: Wow O____o crazy how simple that is

mellogarrett:
Performance is okay. There’s a slight lag opening the list.

boblauer:
Is the lag from the fuzzy searching itself, or perhaps a rendering issue?

mellogarrett:
Not sure to be honest. Very likely it could be a rendering issue. It's not bad at all, just noticeable.

PR #178, opened 2019-08-30
require "great" and "requested_improvements" questions

mellogarrett:


boblauer:
We should merge and deploy this pull to production first before merging the atlas-db pull. If we merge them at the same time, it's possible there will be a period where it's not required on the form but is required on the back end (for example, if someone opened the form this morning but doesn't get around to filling it out until now.)

I would wait probably an hour or so after this goes live before also making it required on the api side.

mellogarrett:
Got it

PR #184, opened 2019-09-05
[CORE-154] Add Search question type

mellogarrett:
- Add QuestionType.Search
- Add search only option to DropdownQuestion
- Adjust Fuse search sensitivity to suggest closer choices.


components/dropdown-question/index.tsx
```diff
@@ -45,7 +45,8 @@ export const DropdownQuestionComponent = ({ question, onChange, style, className
         selectedItem,
         clearSelection,
       }) => {
-        const suggestions = inputValue ? fuse.search(inputValue) : choices;
+        const hasSearchText = inputValue !== null && inputValue !== '';
+        const suggestions = hasSearchText ? fuse.search(inputValue as string).slice(0, 20) : choices;
```
mellogarrett: TypeScript won't infer the type of `inputValue` here as `string` based just on the line above ☹️.

mellogarrett:
This component is getting pretty bloated by adding this `searchOnly` mode. It will get even worse once we add support for multiselect. 

Is there a clean way to break this into two separate components, but still reuse the majority of the markup code? Maybe the answer is simply just have mostly duplicate code in two component files but I'm hesitant of that.

mellogarrett:
The more I look at this, I think what I have is the right approach; keep these 3 use-cases as one component with small bits of conditional logic throughout depending on `QuestionType`.

components/dropdown-question/index.tsx
```diff
@@ -7,112 +7,126 @@ import MenuItem from '@material-ui/core/MenuItem';
...
+  useEffect(() => {
+    setInitialSelectedItem(question.choices.find((choice) => choice.checked));
+  }, []);
+  const isSearchOnly = question.type === QuestionType.Search;
   const { title, description, isRequired, choices } = question;
-  const selectedChoice = question.choices.find((choice) => choice.checked);
   const { suggestionsCss, suggestionControlsCss, suggestionControlsDividerCss, menuItemCss } = generateStyles();
-
-  const fuse = new Fuse(choices, { keys: ['display', 'value'] });
+  const fuse = new Fuse(choices, { keys: ['display', 'value'], threshold: 0.3 });
```
MarkDKoslow: 🎉 Very cool how easy this is to adjust

components/dropdown-question/index.tsx
```diff
@@ -7,112 +7,126 @@ import MenuItem from '@material-ui/core/MenuItem';
...
+                            {...getItemProps({
+                              item: suggestion,
+                              selected: highlightedIndex === i,
+                            })}
+                            css={menuItemCss}
+                          >
+                            <ListItemText primary={display || value} secondary={detail} />
+                          </MenuItem>
+                        );
+                      })
```
MarkDKoslow: Not a blocker 🛠 If the user is seeing 20 suggestions and has scrolled to the bottom of the list, I think we should add a disabled menu item at the bottom that says "Don't see what you're looking for? Continue typing to further refine your search." Otherwise they may think the component is broken or their item isn't in the list.

Maybe we can come back to this next week.

mellogarrett: Oh right. Probably not super common, but might be worth adding later. 

You wanna go ahead and throw a Jira ticket in the backlog for that?

MarkDKoslow: Yeah will do 👍

components/dropdown-question/index.tsx
mellogarrett: agreed.

PR #188, opened 2019-09-05
CORE-151 prepare for server-side rendering and atlas-db

josh-egan:
Requirements this PR addresses:

- Need to prepare for server-side rendering. Server-side rendered forms will have access to `atlas-db` in `getInitialProps`.
- Need to have a custom api so that when a form is submitted, the api can use `atlas-db` to write data.

Solution: next.js 9 :tada:

This turned out to be an easier solution than I expected. Upgrading to next 9 and using next to run the server (see updated "start" script in package.json) will address our requirements.

- The pages are now rendering server-side. The pages are still using `atlas-api`, but `getInitialProps` is now being called server-side only. The existing forms or new forms can start using `atlas-db` in `getInitialProps`.
- Next.js 9 supports a custom api out of the box, so no need to create a custom express server. I haven't made any api routes yet, but when the new forms are developed, or existing forms are converted to use `atlas-db`, an api route can be created using a file at `./pages/api/my-fancy-api-route.ts`. See the [release notes](https://nextjs.org/blog/next-9#api-routes) or [the docs](https://nextjs.org/docs#api-routes) for details.

Some changes I want to call out:

- The previous build command included `cp _redirects out/_redirects` which I assume is something to do with static sites. I removed it - if it needs to be there, let me know.
- I turned off email notifications from travis, because I don't like getting them. If you like email notifications from travis, let me know and I can turn them back on. (See `.travis.yml`)
- The `<Container>` component is deprecated, which is why I removed it from `./pages/_app.jsx`. See  https://github.com/zeit/next.js/blob/master/errors/app-container-deprecated.md If there was a need for the componenet that I'm not aware of, let me know.
- For some reason the loading spinner was breaking, and we couldn't figure out why. Garrett helped me look into it, and we eventually decided it would be most pragmatic to just rip out the material ui component that was breaking and replace it with our own. It sounds like the loading spinner might get ripped out soon anyways when we move to using `atlas-db`.
- The `./pages` folder needed to be re-structured. In next.js 9.x, all of the files in the `.pages` folder will render a page recursively, so it is now possible to created nested paths. Because of that, every file in the pages folder must export a react component and render a page (unless it's in the `api` folder, in which case those files must export an api route handler).

Prior to merging this we need to:

1. [x] Be prepared to freeze forms deploys from other branches until this switch-over is complete.
2. [x] Turn off automatic deploys to netlify for forms staging and production.
3. [x] Create EnvKey configs
3. [x] Setup Heroku builds for staging and prod for forms.

After merging this we need to:

1. [ ] Switch cloudflare staging to point to Heroku.
2. [ ] Test out the forms app in staging and make sure everything is working.
3. [ ] Deploy to production by merging `staging` into `master`
4. [ ] Switch cloudflare production to point to Heroku.
5. [ ] Test out the forms app in production and make sure everything is working.
6. [ ] Tear down the netlify builds.
7. [ ] Remove the `netlify-*` branches from the forms app.


pages/_document.jsx
```diff
@@ -43,6 +51,8 @@ MyDocument.getInitialProps = (ctx) => {
   // 3. app.render
   // 4. page.render
 
+  redirectIfNeeded(ctx);
```
boblauer: Do we not need to `return` here if we have redirected?

next-env.d.ts
```diff
@@ -0,0 +1,2 @@
+/// <reference types="next" />
+/// <reference types="next/types/global" />
```
mellogarrett: What are these for?

tsconfig.json
```diff
@@ -3,7 +3,10 @@
     "allowJs": true,
     "allowSyntheticDefaultImports": true,
     "jsx": "preserve",
-    "lib": ["dom", "es2017"],
+    "lib": [
```
boblauer: This is a note to myself as much as anything, but we should really move this out to our `eslint-config` file.

pages/_document.jsx
```diff
@@ -81,5 +91,20 @@ MyDocument.getInitialProps = (ctx) => {
...
+
+function redirect(ctx, path) {
+  if (ctx.res) {
+    ctx.res.writeHead(302, { Location: path });
+    ctx.res.end();
+  } else {
+    document.location.pathname = path;
+  }
+}
+
```
mellogarrett: Nice!

I wonder if the Next server is particular about having the trailing slash or not?

tsconfig.json
```diff
@@ -14,7 +17,19 @@
...
+    "target": "esnext",
+    "forceConsistentCasingInFileNames": true,
+    "esModuleInterop": true,
+    "resolveJsonModule": true,
+    "isolatedModules": true
   },
-  "exclude": ["out", ".next"]
+  "exclude": [
+    "out",
+    ".next"
```
mellogarrett: Are `out` and `.next` even used with v9?

pages/_document.jsx
josh-egan: It is kind of weird, but we don't need to. `next.js` knows that once the response object has been ended (in the redirect function) that nothing else goes through. I didn't want to return a boolean from that function, which is why I wrote it that way, but it is weird logically. Maybe I could change it to something like

```ts
const redirectPath = getRedirectPath(ctx);
if (redirectPath) return redirect(ctx, redirectPath);
```

tsconfig.json
josh-egan: `.next` is - that's where `npm run build` appears to dump the built files. Doesn't look like `out` is used anymore.

next-env.d.ts
josh-egan: `next` added them automatically.

pages/_document.jsx
josh-egan: It's not particular. I tried both `/sprint-retro` and `/sprint-retro/` and they both work fine. :trophy:

josh-egan:
FYI I'm still working on getting the tests to pass. jest seems to be having a struggle locating elements it could apparently find before.

components/loading-screen/index.tsx
```diff
@@ -8,19 +7,84 @@ export const LoadingScreen = () => {
...
 const generateStyles: StyleGenerator = () => {
+  // Spinner courtesy of https://projects.lukehaas.me/css-loaders/
   return {
     loadingScreenStyles: css`
       height: 100vh;
       width: 100vw;
       display: flex;
       justify-content: center;
       align-items: center;
+
```
MarkDKoslow: Thanks for switching this 👏

pages/_document.jsx
```diff
@@ -81,5 +92,21 @@ MyDocument.getInitialProps = (ctx) => {
   };
 };
 
+function getRedirectPath(ctx) {
+  const p = ctx.asPath || '';
+  if (p === '/') return `https://dashboards.${getDomain()}/forms`;
```
MarkDKoslow: Note that we should update dashboards - `/forms` works but `/forms/` does not

PR #189, opened 2019-09-05
Update description to tell students to not use their Build Week TL

MarkDKoslow:


MarkDKoslow:
Just caught up with Patrick and Ryan Hamblin and they mentioned that they still want students to be able to review their Project Lead (instead of TL), so I'm going to close this. Now that TL's can review any sprint retro, I don't think students necessarily need to select their TL

PR #202, opened 2019-09-13
SSR fixes

mellogarrett:
I was wrong about how much of the `_app` and `_document` we could safely remove 😅


mellogarrett:
For instance, go to https://forms.lambdaschool.com/test-page to see the issues.

PR #204, opened 2019-09-16
Job search check in client

MarkDKoslow:
Add form data and backend URL's for Job Search Check In form


forms/job-search-check-in.ts
```diff
@@ -0,0 +1,292 @@
+import { Form, Backend, QuestionType } from '../types';
+import { FormTitle } from '../constants/form-title';
+
+export const JobSearchCheckInForm: Form = {
+  title: FormTitle.JobSearchCheckIn,
+  description:
+    'Sample description here',
```
MarkDKoslow: Going to be updated

forms/job-search-check-in.ts
```diff
@@ -0,0 +1,291 @@
...
+          display: '6+',
+          value: '6+',
+          checked: false,
+        },
+      ],
+      externalKey: 'onsite_interviews',
+      hidden: false,
+    },
+    {
+      title: 'How many meetups or events did you plan or attend this week?',
```
brentjanderson: Is this right? `Plan or attend?` Are students planning meetups or events?

http/submit-form/backend-adapters/zapier/zapier-urls.ts
```diff
@@ -17,6 +17,7 @@ const zapierDevelopmentUrls = {
   [FormTitle.TestForm]: '',
   [FormTitle.TLSprintReview]: 'https://hooks.zapier.com/hooks/catch/4903492/oot4r4n/',
```
brentjanderson: Not among your changes, but is there a reason we don't make these values into environment variables?

types/questions.ts
```diff
@@ -74,6 +74,14 @@ interface BaseQuestion {
     | 'student_entered_instructor'
     | 'student_entered_pm'
     | 'section_id'
+    | 'applications_submitted'
```
brentjanderson: Would we want to extend `BaseQuestion` for each form so that we have the right keys available per form, but not external keys from other forms?

forms/job-search-check-in.ts
```diff
@@ -0,0 +1,291 @@
+import { Form, Backend, QuestionType } from '../types';
+import { FormTitle } from '../constants/form-title';
+
+export const JobSearchCheckInForm: Form = {
```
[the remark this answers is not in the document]
mellogarrett: It’d be pretty straightforward I think

http/submit-form/backend-adapters/zapier/zapier-urls.ts
mellogarrett: We have several url maps throughout the app. I’m not sure moving them to environment variables buys us anything. It would just be managed in Heroku rather than in our code base.

forms/job-search-check-in.ts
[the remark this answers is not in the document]
brentjanderson: Totally out of scope for this, but I'll drop a Jira ticket for it to document it.

types/questions.ts
mellogarrett: Not sure what you mean here. Care to elaborate? Right now the different question types extend BaseQuestion

brentjanderson: This is a long list of external keys. Would it make more sense to do something like 

```typescript
interface JobSearchCheckinQuestions extends BaseQuestion {
externalKey: 'keys' | 'go' | 'here'
}
```

Right now there's a long list of each external key, and presumably each form will care about its own external keys rather than sharing them across all the forms? This is a nitpick, so I'm fine either way on this, but it seems a bit odd having an ever-growing list of keys that would presumably belong to only one form for each key, or if a key is shared between forms it could be copied in both places for each form to have its own list of accepted external keys.

mellogarrett: Oh I see. Yea I like that idea.

forms/job-search-check-in.ts
```diff
@@ -0,0 +1,291 @@
...
+  title: FormTitle.JobSearchCheckIn,
+  description: 'Sample description here',
```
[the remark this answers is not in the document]
MarkDKoslow: Good call, now updated 👍

forms/job-search-check-in.ts
MarkDKoslow: I think so according to the jira ticket - `Meetups or Events planned or attended`

types/questions.ts
MarkDKoslow: Yeah I like that as well. I'll add a ticket to come back to this

MarkDKoslow: https://lambdaschool.atlassian.net/browse/CORE-245?utm_medium=referral-external&utm_source=jira-slack&utm_term=

PR #206, opened 2019-09-17
[CORE-264] Server-rendered Job Search Check In Form

mellogarrett:
There's a lot going on here. I tried to include as much detail in the commit messages as possible. Locally this is pre-populating "Your Name" options correctly on the server.

I still need to write/fix some tests but wanted to get this out here for you all to have a look.

- Custom server to ensure the `Instruction` context has `.connect()`ed before routes start taking requests
- Handles pre-populating from both PostgREST and Forms API (snakecase and camelcase option keys & values)
- `prepopulateQuestions` has been updated so that it is "isomorphic" or can be ran on the server or the client. This means I didn't make `atlas-db` calls directly here, I went ahead and used `/GET` endpoint Josh & Mark created and am calling with an Axios client. There were a couple things I avoided by doing this that would've added complexity:
  - no need to worry about mocking `atlas-db` packages in the client with Webpack
  - the `req/res` objects available in `.getInitalProps` are different than the `req/res` objects available in `/api` routes meaning we'd need two different versions of the `getUserFromJwt` and `redirectToLogin` functions; one for client-side and one for server-side
- Since we aren't using `atlasClient` in `prepopulate-questions` and are instead using a plain Axios client, I updated the `get-prepopulate-url.ts` function to be a map of exact URLs rather than relative URLs. This means we can have Forms API & PostgREST prepopulate URLs co-existing here.
- When server-rendering, we can't redirect `302` to the login page so instead I'm returning `401` and then rendering a new `<Unauthorized/>` page instead with a link to the login page.


pages/api/job-search-check-in.ts
```diff
@@ -0,0 +1,27 @@
...
+import { SubmitJobSearchCheckInFormParams } from '@lambdaschool/atlas-db/dist/contexts/instruction';
+import { atlas } from '../../domain/atlas';
+import { getUserFromJwt, redirectToLogin } from '../../middleware';
+
+export default async (req: NextApiRequest, res: NextApiResponse) => {
+  try {
+    const user = await getUserFromJwt(req);
+    if (!user) return redirectToLogin(req, res);
+
+    if (req.method === 'POST') return submitForm(user.id, req.body).then(() => res.status(200).end());
```
boblauer: The status here should be `201`, which is `Created`.

pages/api/job-search-check-in.ts
```diff
@@ -0,0 +1,27 @@
...
+
+    res.status(404).end();
+  } catch (e) {
+    res.status(500).end();
+  }
+};
+
+async function submitForm(userId: string, data: object) {
+  const postData = (data as unknown) as SubmitJobSearchCheckInFormParams;
+  await atlas.instruction.submitJobSearchCheckInForm(userId, postData);
```
boblauer: There's no reason to `await` here, it's the same as `return atlas...`.

pages/api/job-search-check-in.ts
```diff
@@ -0,0 +1,27 @@
...
+}
+
+async function handleGetRequest(userId: string) {
+  return (await atlas.instruction.getJobSearchCheckInForm(userId)) as object;
```
boblauer: No reason to `return await` here, `return atlas...` achieves the same thing without the indirection.

types/forms-api.ts
```diff
@@ -0,0 +1,14 @@
+export interface FormsApiOption {
+  questionId: string;
+  optionKey: string;
+  optionValue: string;
+  optionDetail?: string;
+}
+
+export function isFormsApiOption(x: any): x is FormsApiOption {
+  return Object.keys(x).some((key) => {
```
boblauer: Having a `some` with a `? false : true` after it hurts my brain. I think this is equivalent to

```js
return Object.keys(x).all(key => ['questionId', 'optionKey', 'optionValue', 'optionDetail'].includes(key));
```

http/prepopulate-questions/index.test.ts
```diff
@@ -1,22 +1,27 @@
...
 import { prepopulateQuestions } from '.';
 import * as GroupOptions from '../../utils/group-options';
 import * as UpdateQuestions from './update-questions';
 import { mockForm } from '../../__mocks__/form.mock';
 import { mockAtlasGet } from '../../__mocks__/atlas-get.mock';
-import { mockCreateAtlasClient } from '../../__mocks__/create-atlas-client.mock';
 import { mockLogger } from '../../__mocks__/logger.mock';
+import axios from 'axios';
+import * as getPrepopulateUrl from '../../utils/get-prepopulate-url';
+import * as getDomain from '@lambdaschool/package-utils';
```
boblauer: I would probably name this `as packageUtils` rather than `as getDomain`, since it's the full object and not just the `getDomain` function.

package.json
```diff
@@ -31,13 +32,15 @@
...
     "fuse.js": "3.4.5",
+    "jsonwebtoken": "8.5.1",
     "linkifyjs": "2.1.8",
     "lodash.isequal": "4.5.0",
     "next": "9.0.5",
     "react": "16.9.0",
     "react-dom": "16.9.0",
     "react-ga": "2.5.7",
     "react-jss": "8.6.1",
+    "ts-node": "8.4.1",
```
boblauer: Dev dependency?

pages/daily-standup.tsx
```diff
@@ -4,5 +4,4 @@ import { DailyStandupForm } from '../forms';
 
 const DailyStandupFormPage = () => <FormComponent pageTitle="Standup Form | Lambda School" form={DailyStandupForm} />;
 
-/* eslint-disable-next-line import/no-default-export */
```
boblauer: I'm curious, why this is no longer necessary?

utils/get-forms-api-url.ts/index.ts
```diff
@@ -0,0 +1,12 @@
+import { FormTitle } from '../../constants/form-title';
+
+export const getFormsApiUrl = (formTitle: FormTitle) =>
```
boblauer: Is this file used anywhere?

pages/api/job-search-check-in.ts
mellogarrett: We decided we don't want to return the created record back to the client here. We just want to wait until it's created and then return nothing.

boblauer:
Lots of linting errors. You may want to configured your IDE to scream at you when you have a lint error, to avoid having to clean up after the fact.

pages/api/job-search-check-in.ts
boblauer: We're ignoring the returned promise's value on L11 above, meaning the return value here won't affect the response we send to the server.

That said, if we really don't want to return anything from this function, can we add a `return null` after this to make it explicit?

server/index.ts
```diff
@@ -0,0 +1,22 @@
+import { createServer } from 'http';
+import { parse } from 'url';
+import next from 'next';
+import { atlas } from '../domain/atlas';
+
+const portArg = process.argv[process.argv.findIndex((arg) => arg === '-p') + 1];
```
MarkDKoslow: Ah this is clever

mellogarrett:
Agreed Mark. The added conditional stuff kinda stinks but I think we’ll be able to rethink it a bit once we migrate off like you mentioned.

http/submit-form/backend-adapters/forms-api/index.ts
```diff
@@ -0,0 +1,38 @@
...
+
+const formatAtlasRequest = (form: Form, mergeId: string) => {
+  const formSubmission: Record<string, Answer> = {};
+
+  if (form.backends.includes(Backend.Zapier)) {
+    // eslint-disable-next-line @typescript-eslint/camelcase
+    formSubmission.merge_id = mergeId;
+  }
+
+  return form.questions.reduce((submission, question) => {
```
boblauer: Looks like if there's no `externalKey` then we do nothing here, in which case we can break that apart into two steps:

```js
return form.questions
  .filter(question => question.externalKey)
  .reduce((submission, question) => Object.assign(submission, {
    [question.externalKey]: getAnswerForQuestion(question)
  }), formSubmission);
```

mellogarrett:
True, although that means we'd be iterating over the same array twice which seems unnecessary to me.

boblauer:
I can't imagine we ever have more than 30 questions for a given form. Even if we had 1,000 questions, iterating over that array twice instead of once would maybe add 1ms, _maybe_.

I would always opt for smaller functions that do one thing over micro optimizations.

http/submit-form/backend-adapters/forms-api/index.ts
mellogarrett: Actually, it looks like we can just remove that line. A `Question` will always have an `externalKey`.

mellogarrett:
Alright, I believe this in a state to merge now. Any final reviews would be appreciated. 

I've tested locally and verified that authenticating, prepopulating, and submitting works on both client & server-rendered pages. It would be nice to do a similar test when this is deployed on Heroku.

pages/job-search-check-in-server-rendered.tsx
```diff
@@ -0,0 +1,20 @@
...
+import { serverSidePrepopulateQuestions } from '../http/server-side-prepopulate-questions';
+
+const JobSearchCheckInPage = ({ unauthorized, form, redirectAfterLoginUrl }: ServerRenderedPageProps) =>
+  unauthorized ? (
+    <UnauthorizedComponent redirectAfterLoginUrl={redirectAfterLoginUrl} />
+  ) : form ? (
+    <FormComponent pageTitle="Job Search Check In | Lambda School" form={form} noResolver={true} />
+  ) : null;
+
+JobSearchCheckInPage.getInitialProps = async (ctx: NextPageContext) =>
```
boblauer: Can do we something like this to achieve a 401 redirect?

```js
JobSearchCheckInPage.getInitialProps = async (ctx: NextPageContext) => 
{
	const props = await serverSidePrepopulateQuestions(JobSearchCheckInForm, ctx);
	if (!props.unauthorized) return props;

	res.writeHead(302, {
		Location: props.redirectAfterLoginUrl
	});
	res.end();
});
```

mellogarrett: Possibly, let me play with this.

mellogarrett: Alright, I was able to get this working 🎉

mellogarrett:
Made a pretty substantial refactor to the `http/prepopulate-questions/` folder (and also renamed it to `merge-question-options`) to get rid of all the conditional logic. This is now split out by backend.

Also, the redirect when unauthorized is working now so I removed the `<UnauthorizedComponent/>`

mellogarrett:
Agreed, sorry about that.

PR #207, opened 2019-09-18
Job search checkin remove banner

MarkDKoslow:
Namrata asked if we could remove the banner for the Job Search Check In form. I think the easiest way to do this is to include an optional list of FormOptions for each form. Right now the options are essentially restricted to boolean values, but we could expand this in the future if we want to have multiple variants of a given "option".


boblauer:
fwiw I agree this probably isn't the ideal solution, and I think it will likely be refactored in the future. But if this is something we want out asap, I'm good with merging this and then refactoring in a separate pull.

mellogarrett:
Agreed. I’m fine with merging as is for now.

MarkDKoslow:
Okay I'm going to go ahead and merge this and then we can refactor afterwards

PR #214, opened 2019-09-23
CORE-245 Break up long list of "externalKey"s

mellogarrett:
## Jira Issue

https://lambdaschool.atlassian.net/browse/CORE-245?atlOrigin=eyJpIjoiMDQzYTA0YTk3YTdjNDIxN2I0ZTM3YThkMjdhYTJjZjUiLCJwIjoiaiJ9

## Description

- Broke out `externalKey` values by form
- Deleted unused forms
- Fixed bug where page title would be "My Page" on initial render and after submitting your responses.

## Type of change


types/questions.ts
```diff
@@ -35,62 +42,10 @@ interface BaseQuestion {
...
-    | 'studentId'
-    | 'applications_submitted'
-    | 'phone_screens'
-    | 'technical_interviews'
-    | 'onsite_interviews'
-    | 'events_attended'
-    | 'networking_hours'
-    | 'offers_received'
-    | '';
+    | SprintRetroExternalKeys
```
brentjanderson: Very nice!

forms/test-page.ts
```diff
@@ -12,7 +12,7 @@ export const TestForm: Form = {
       isRequired: true,
       prepopulateChoices: true,
       choices: [],
-      externalKey: 'student_id',
+      externalKey: 'student_id' as any,
```
brentjanderson: I'm seeing a lot of `any` in this PR - is it required because of the changes to the `externalKey` types below? What would be a way to avoid peppering the code with `any` when we reference externalKey?

mellogarrett: All of the `any`s should be related to test code only. We could use legitimate values for these but since the keys are really determined outside of this app and might change, it felt okay to just use a bogus key with ` as any` in tests.

pages/_app.jsx
```diff
@@ -38,7 +38,7 @@ class MyApp extends App {
     return (
       <Fragment>
         <Head>
-          <title>My page</title>
+          <title>Lambda Forms</title>
```
MarkDKoslow: 👏

types/external-keys.ts
```diff
@@ -0,0 +1,62 @@
+export type JobSearchCheckInExternalKeys =
+  | 'student_id'
+  | 'escalations'
```
MarkDKoslow: Hmm trying to think through the use cases here on whether it'd be more helpful if these are chronological or alphabetical... I'm thinking alphabetical because we'll likely have a key in mind if we want to edit the existing list, but certainly not a blocker

PR #228, opened 2019-09-26
CORE-339 Refactor "validators"

mellogarrett:
## Jira Issue

https://lambdaschool.atlassian.net/browse/CORE-339?atlOrigin=eyJpIjoiNDc0NjJmYzQyMjE5NGQ0NGIyNjJlYjFhMmMwOWI2NGQiLCJwIjoiaiJ9

## Description

Server-rendering was breaking validations because the validation functions were being serialized out of the form object in `getInitialProps`. 

This works around that by using an enum of possible validators, then getting that validator function on the client when needed.

## Type of change
_Place an `x` in between the corresponding `[ ]`._


components/number-question/index.test.tsx
```diff
@@ -15,6 +20,15 @@ const question: NumberQuestion = {
 };
 
 describe('NumberQuestion', () => {
+  beforeAll(() => {
+    jest.spyOn(getValidatorFunction, 'getValidatorFunction').mockImplementation((validator: string) => {
+      const map: { [key: string]: Validation } = {
```
boblauer: Is this worth doing? We know we'll always be returning `lengthValidation`, could we instead just do something like

```
jest.spyOn(...).mockImplementation(() => lengthValidation)
```

I think that'll let us get rid of the `enum Validator` up top, and then on L70 below just do `validators: ['length'] as any`.

mellogarrett: Yea, definitely not worth doing in this instance. I was just copying & pasting that pattern from test to test. 

I'll fix.

components/number-question/index.test.tsx
```diff
@@ -53,7 +67,7 @@ describe('NumberQuestion', () => {
     const questionWithError: NumberQuestion = {
       ...question,
       value: 1,
-      validators: [validation],
+      validators: [Validator.length] as any,
```
mellogarrett: Hmm, if we refactored the enum out this would then just be some random string. 

I kinda like using the full enum pattern here, even if it's a bit more code in this case, just because it better represents how validators are actually used. 

Thoughts @boblauer ?

boblauer: I think the ideal scenario is that we don't have to mock validation at all. However, my guess is that we do mock it because for some of these questions, we don't have an existing validator that makes sense to test.

One idea is to get rid of `isRequired` and move that into a validator. That would allow us to use that validator in all of our tests and simplify the code a bit by consolidating field requirement into the validator logic.

All of that said, not necessary to do that in this pull request. I'm good with leaving the enum in for now.

PR #242, opened 2019-10-04
[CORE-162] Add support for selecting multiple options to Dropdown

mattbspector:
@MarkDKoslow @mellogarrett 

Please *link* to the corresponding Jira issue.

https://lambdaschool.atlassian.net/browse/CORE-162

## Description

Adding support for users to select multiple options from the dropdown list

image

 
## Type of change


components/dropdown-question/index.tsx
```diff
@@ -1,11 +1,12 @@
 import React, { CSSProperties, ChangeEvent, useMemo } from 'react';
-import { MenuList, IconButton, Divider, ListItemText } from '@material-ui/core';
+import { MenuList, IconButton, Divider, ListItemText, Icon } from '@material-ui/core';
 import { css } from '@emotion/core';
 import TextField from '@material-ui/core/TextField';
 import Paper from '@material-ui/core/Paper';
 import MenuItem from '@material-ui/core/MenuItem';
 import Downshift from 'downshift';
-import { Close, ExpandMore } from '@material-ui/icons';
+import ChipInput from 'material-ui-chip-input';
```
mellogarrett: There should be a `@material-ui/core/Chip` component that we can use instead of this new dependency.

components/question/index.tsx
```diff
@@ -49,8 +49,11 @@ export const QuestionComponent = ({ question, questionIndex, ...rest }: Question
     }
 
     switch (question.type) {
-      case QuestionType.Dropdown:
       case QuestionType.Search:
+      case QuestionType.Dropdown:
+        return question.multipleSelect
+          ? dispatch(updateMultiOptionAction(value, questionIndex))
+          : dispatch(updateSingleOptionAction(value, questionIndex));
```
mellogarrett: Nice. I think I would alter the switch statement to take the whole question object now though, since we're checking multiple properties on it.

components/dropdown-question/index.tsx
```diff
@@ -112,6 +126,11 @@ export const DropdownQuestionComponent = ({ question, onChange, style, className
                             css={menuItemCss}
                           >
                             <ListItemText primary={display || value} secondary={detail} />
+                            {checked ? (
+                              <Icon css={listIconCss} aria-label="option selected">
+                                <Star />
+                              </Icon>
+                            ) : null}
```
mellogarrett: Hmm, this is just my preference but I wonder if instead of adding the star at the end, we could just have selected items have a dark background. 

Curious to get others' thoughts here.

components/dropdown-question/index.tsx
```diff
@@ -59,48 +60,61 @@ export const DropdownQuestionComponent = ({ question, onChange, style, className
...
-                      )}
-                    </div>
-                  ),
-                })}
-                placeholder="Search..."
-              />
+              {multipleSelect ? (
+                <ChipInput
+                  fullWidth
+                  value={choices.filter((choice) => !!choice.checked).map((choice) => choice.display)}
```
mellogarrett: You'll need to fallback to `choice.value` if `choice.display` isn't defined here since `.display` is optional.

components/dropdown-question/index.tsx
```diff
@@ -44,7 +45,7 @@ export const DropdownQuestionComponent = ({ question, onChange, style, className
           const event: any = { target: { value: item ? item.value : null } };
           onChange(event);
         }}
-        itemToString={(item) => (item ? item.display || item.value : '')}
+        itemToString={(item) => (multipleSelect ? '' : item ? item.display || item.value : '')}
```
mellogarrett: Curious why we need this check here?

components/dropdown-question/index.tsx
```diff
@@ -19,9 +20,9 @@ interface DropdownQuestionProps {
 }
 
 export const DropdownQuestionComponent = ({ question, onChange, style, className }: DropdownQuestionProps) => {
-  const { title, description, isRequired, choices } = question;
+  const { title, description, isRequired, choices, multipleSelect } = question;
```
mellogarrett: Maybe just name this `isMultipleSelect` or something similar to follow the convention.

utils/get-dropdown-answer/index.ts
```diff
@@ -1,7 +1,12 @@
 import { DropdownQuestion, Answer, SearchQuestion } from '../../types';
 
 export const getDropdownAnswer = (question: DropdownQuestion | SearchQuestion): Answer => {
-  const checkedChoice = question.choices.find((choice) => choice.checked);
+  const isMultiDropdown = !!question.multipleSelect;
+
+  if (isMultiDropdown) {
+    return question.choices.filter((choice) => !!choice.checked).map((choice) => choice.value);
+  }
```
mellogarrett: I don't think you need `isMultiDropdown` on line 4. you can just do `if (question.multipleSelect)` on line 6. 

Also, the `!!` on `!!choice.checked` shouldn't be needed.

components/dropdown-question/index.tsx
mattbspector: So the reason I added this package is because it gives us the text field input AND the chips bundled together. Otherwise we'd basically have to write what they have written.

mellogarrett: Ah okay. Do you think it might limit our flexibility over just building them together ourselves?

mattbspector: I think it might be flexible enough to support what we need: https://www.npmjs.com/package/material-ui-chip-input but I'll look into building it ourselves now.

components/question/index.tsx
mattbspector: 👍good call

components/dropdown-question/index.tsx
mattbspector: I definitely agree, will change that up now

components/dropdown-question/index.tsx
mattbspector: 👍

components/dropdown-question/index.tsx
mattbspector: If we don't have this check, after you search -> select an option, it will leave that value in the text field. This makes it so that the text field clears after selection

components/dropdown-question/index.tsx
mattbspector: 👍

components/dropdown-question/index.tsx
MarkDKoslow: Hmm I'm not sure a dark background would convey something being selected without some sort of check mark or symbol

mattbspector: We can do both! I'll change the star to a checkmark and highlight the row as well

mellogarrett: That's how the Material UI examples for multi-select look and they work well in my eyes.

https://material-ui.com/components/selects/#multiple-select

mattbspector:
@mellogarrett @MarkDKoslow Added tests and my package-lock file! 😄

components/dropdown-question/index.tsx
MarkDKoslow: Oh i see, I was thinking it was just making that background slightly darker, but that example makes it more clear. I'm onboard for any of the above.

mattbspector:
@MarkDKoslow @mellogarrett Added the different colors for hover and made the chip labels use the chip value!

mattbspector:
Also added a test for that behavior @MarkDKoslow

PR #262, opened 2019-10-10
[Core-411] Finish work for peer review form

MarkDKoslow:
## Jira Issue
[[CORE-411]](https://lambdaschool.atlassian.net/browse/CORE-411)

## Description
Wrap up the FE and api work for Peer Review form

Just need to wait for backend deploy to finish and then update the atlas-db package before merging.

## Type of change


http/submit-form/backend-adapters/forms-api/forms-api-urls.ts
```diff
@@ -16,5 +16,5 @@ export const getFormsApiUrl = (formTitle: FormTitle) =>
     [FormTitle.JobSearchCheckIn]: `https://forms.${getDomain()}/api/job-search-check-in`,
     [FormTitle.UnitSubmission]: `https://forms.${getDomain()}/api/unit-submission`,
     [FormTitle.UnitSubmissionReview]: `https://forms.${getDomain()}/api/unit-submission-review`,
-    [FormTitle.PeerReview]: '',
+    [FormTitle.PeerReview]: `https://forms.${getDomain()}/api/peer-review`,
```
brentjanderson: Comment: Wouldn't it be great if we just used lodash `kebabCase` on the properties of `FormTitle` and generated the URLs from that? Any reason not to adopt a convention like that?

pages/api/peer-review.ts
```diff
@@ -0,0 +1,31 @@
...
+    if (!user) {
+      return res.status(401).end();
+    }
+
+    if (req.method === 'POST') return submitForm(user.id, req.body).then(() => res.status(201).end());
+    if (req.method === 'GET') return getFormData(user.id).then((r) => res.json(r));
+
+    res.status(404).end();
+  } catch (err) {
+    console.error(err);
```
brentjanderson: Do we not have `@lambdaschool/logger` setup for rollbar to pick stuff like this up?

http/submit-form/backend-adapters/forms-api/forms-api-urls.ts
mellogarrett: It would tightly couple us to the form name here, which is slightly undesirable in my mind.

pages/api/peer-review.ts
mellogarrett: We should definitely do that in a future PR.

http/submit-form/backend-adapters/forms-api/forms-api-urls.ts
MarkDKoslow: Yeah would agree with Garrett, especially given our fondness for name changes as an org 🙂

pages/api/peer-review.ts
MarkDKoslow: Yeah good call

PR #263, opened 2019-10-10
Multi select escalation

brentjanderson:
## Jira Issue

N/A

## Description

The spec calls for the escalation to be multi-select, and confirmed with Aparna as such.

## Type of change


brentjanderson:
> Hey Brent, will this question be updated to multi-select for all of the new forms that include this question? I believe there's 4 of them.

I can't say. All I know is that the peer review form was specifically intended to have a multi-select escalation field, so that's why it's designed this way. The other forms would need a nod from product to know for sure.

PR #266, opened 2019-10-11
CORE-121 implement local jwt decoding in forms app

josh-egan:
## Jira Issue

[CORE-121]

## Description

Implement local JWT decoding using the new public-keys endpoint in the auth app.

I am going to extract this functionality into `package-utils` soon. Now that I've implemented this in dashboards and forms, the way to make this functionality reusable is pretty clear. After it has been extracted, I'll come back through and update the forms app with functions from `package-utils`.

Some notes:
- Upgraded node version. The `jose` package, which does the JWT decoding, requires node >= 10.13. The dashboards project had several errors when I tried to upgrade to node 12, so I went with the most recent version of 10 instead. I have upgraded this project to the same version as I did dashboards.
- `NODE_TLS_REJECT_UNAUTHORIZED` is now required for local dev only and has been added to the .env.example file.

I'm going to tag a bunch of people for information sharing.

## Type of change

[CORE-121]: https://lambdaschool.atlassian.net/browse/CORE-121


server/utils/authentication.ts
```diff
@@ -14,20 +19,79 @@ interface LambdaUser {
...
+
+  if (req.headers && req.headers.authorization && req.headers.authorization.startsWith('Bearer ')) {
+    return req.headers.authorization.split(' ')[1];
+  }
+
+  return null;
+}
+
+const CACHE_KEY = 'is-key-store-fresh';
+const CACHE_TIMEOUT = 1000; // 30 minutes in ms
```
brentjanderson: Is this 30 minutes in ms?

josh-egan: Ah, great catch!! I had it set to 1 second for testing.

boblauer:
This is only slightly related, as the `1000` issue was a debugging issue and not a genuine mistake, but when I see `1000 * x * y; // z in ms`, I rarely double-check that math.

It might be useful to standardize on something like https://www.npmjs.com/package/unitime to avoid an incorrect formula from causing some event to happen way too often/seldom.

PR #278, opened 2019-10-15
upgrade package-utils and use new functions

josh-egan:
## Jira Issue

[CORE-121]

## Description

- Upgrade to the latest version of package-utils and use its functions
    - verifyEnvVars is used in the /server/index.ts file
    - logBasicEnvInfo is used in the /server/index.ts file
    - getUserFromRequest is used in server/utils/authentication.ts
- The jwt libraries do not work client side, so the package-utils jwt file is excluded in the webpack config.
- The now unused `JWT_SECRET` env variable was removed.

## Type of change

[CORE-121]: https://lambdaschool.atlassian.net/browse/CORE-121


server/utils/authentication.ts
```diff
@@ -1,97 +1,9 @@
...
+import { getUserFromRequest, LambdaUser } from '@lambdaschool/package-utils';
 import { NextApiRequest } from 'next';
-import axios from 'axios';
 import { logger } from '../../utils/logger';
 
 export async function getUserFromJwt(req: NextApiRequest): Promise<LambdaUser | null> {
   const user = await getUserFromRequest(req);
   logger.debug('Retrieved user from jwt', { user });
   return user;
 }
```
mellogarrett: I don't see where we're using this function

josh-egan: The `getUserFromJwt` function is currently used in all of the api routes. For example, https://github.com/LambdaSchoolEngineering/forms/blob/staging/pages/api/sprint-submission.ts#L9

mellogarrett: Ah whoops

next.config.js
```diff
@@ -13,6 +13,10 @@ module.exports = {
       test: /\.(test|spec)\.ts(x?)$/,
       loader: 'ignore-loader',
     });
+    config.module.rules.push({
+      test: /@lambdaschool\/package-utils\/dist\/jwt.js$/,
+      loader: 'ignore-loader',
+    });
```
mellogarrett: Did you run into an issue without this? I thought it would be fine since we don't import that library anywhere in the frontend code.

PR #280, opened 2019-10-15
CORE-735 Add Question Sections

mellogarrett:
## Jira Issue

[CORE-735]

## Description

- Add Question Section groups
- Update Unit Submission form with groups from Aparna

We now have questions located in two places on the form:
1) `form.sections` This is where questions should be defined by us and this is what controls how questions & sections are rendered.
2) `form.questions` This is created on the fly based on what's defined in `form.sections`. This is where the question state lives, like what choices are currently selected, etc. and this is what is used to submit responses to a given backend. 

## Type of change


boblauer:
I'm confused why that would be. Updating the `question` object should update it in both places, since both arrays hold a reference to the same object. A simple example:

```js
const sections = [{
  questions: [{
    name: 'awesome',
    options: []
  }]
}];

const questions = sections.reduce((all, section) => [...all, ...section.questions], []);

questions[0].options.push('value 1');

console.log(sections[0].questions[0].options);
// logs 'value 1'
```

boblauer:
This is actually where having a `BaseForm` class would be useful. We could have an `updateQuestion` method on the `form` object, and coordinate everything in there.

mellogarrett:
Yea I see...

Hmm, maybe splitting these into two separate lists wasn't such a great idea.

mellogarrett:
Closed in favor of #282

PR #282, opened 2019-10-16
CORE-735 Add Question Groups

mellogarrett:
## Jira Issue

[CORE-735]

## Description

- Add Question Groups
- Rename `question.externalKey` to `question.key`
- Update store to update questions based on `question.key` instead of index

## Type of change


mellogarrett:
This is a different approach to implement what we worked on yesterday, but it doesn't required maintaining a separate `form.sections` and `form.questions` array.

This also enables you colocate Questions and QuestionGroups in `form.questions` rather than requiring everything to be a QuestionGroup.

mellogarrett:
This is working functionally for me locally, but no tests have yet been updated.

types/questions.ts
```diff
@@ -119,3 +119,14 @@ export type Question =
   | CheckboxQuestion
   | RatingQuestion
   | RadioQuestion;
+
+export type QuestionGroup = {
+  group: string;
+  questions: Question[];
+};
+
+export function isQuestionGroup(x: Question | QuestionGroup): x is QuestionGroup {
```
boblauer: fwiw, it feels odd for functionality to live inside of the `types` folder. I would expect `types` to only contain type definitions.

mellogarrett: Open to suggestions on where this would go. It's only purpose is to assert that something is of a given type so I figured I would just colocate it with the type itself.

http/submit-form/backend-adapters/atlas/index.ts
```diff
@@ -10,27 +10,34 @@ export const atlasBackendAdapter: BackendAdapter = {
...
     formSubmission.merge_id = mergeId;
   }
 
-  return form.questions
-    .filter((question) => (question.externalKey as any) !== '')
-    .reduce((submission, question) => {
-      const { externalKey } = question;
-      const answer = getAnswerForQuestion(question);
+  form.questions.forEach((questionOrQuestionGroup) => {
+    if (isQuestionGroup(questionOrQuestionGroup)) {
```
boblauer: This is a common pattern throughout, where we're looping over `form.questions` and then if it's a question group, we loop over all of those questions and do X, and if not we loop over the questions directly and do X.

Because these functions all take a `form`, we can't use recursion. But if instead the function took an object with a `questions` property, then we could use recursion. For this specific function, `formatRequest` needs to take a full form because it has to set `merge_id`. But if you pulled that functionality out into a separate function, so that this function only looped over the questions, you could do something like:

```js
function formatRequest(questionContainer: Form | QuestionGroup) {
  const formSubmission = {};

  questionContainer.questions.forEach((questionOrQuestionGroup) => {
    if (isQuestionGroup(questionOrQuestionGroup)) {
      Object.assign(formSubmission, formatRequest(questionOrQuestionGroup));
    } else {
      questionOrQuestionGroup.questions
        .filter(question => question.externalKey)
        .forEach((question) => Object.assign(formSubmission, { [question.externalKey]: getAnswerForQuestion(question));
    }
  });
}
```

mellogarrett: Ah cool

mellogarrett: done

mellogarrett:
Tests updated (hardly any changes 🎉). This is ready to merge once reviewed.

components/form/index.tsx
```diff
@@ -43,15 +43,37 @@ export const FormComponentInner = ({ pageTitle, noResolver }: FormComponentInner
...
+          {questionOrQuestionGroup.group}
+        </Typography>
+        {questionOrQuestionGroup.questions.map((question, questionIndex) => {
+          const groupIndex = questionOrQuestionGroupIndex;
+          return (
+            <QuestionComponent
+              key={question.externalKey}
+              question={question}
+              questionIndex={questionIndex}
+              groupIndex={groupIndex}
```
boblauer: Given that every question now has a unique `externalKey` (is that assumption correct?), updating questions in the reducer by index seems unnecessary. I don't know how big of a change it would be, but we may want to consider passing in the `externalKey` (since I don't think `key` actually gets passed through to `props`) to then use to identify the question that was updated in the action/reducer.

Thoughts?

mellogarrett: Yea I like that idea a lot. Definitely seems doable and yes externalKey should always be unique since it maps to database column names.

mellogarrett: Should I go ahead and make that adjustment in this PR, or do it as a separate PR?

mellogarrett: Just decided to make a branch off of this branch

PR #283, opened 2019-10-17
CORE-735 update questions by key

mellogarrett:
## Jira Issue

[CORE-735]

## Description

An update to CORE-735 where the store now handles updates using the `question.key` value (formerly `externalKey`).

## Type of change


boblauer:
Lots of failing tests around not finding the question object in the reducer.

store/reducers/utils.ts
```diff
@@ -1,34 +1,37 @@
...
+import { Form, QuestionKey, isQuestionGroup, QuestionOrQuestionGroup, Question } from '../../types';
 
 interface UpdateQuestionParams {
   form: Form;
-  action: { questionIndex: number; groupIndex?: number; value: any };
+  action: { key: QuestionKey; value: any };
   updateFunction: (question: any, value: any) => void;
 }
 
 export const updateQuestion = ({ form, action, updateFunction }: UpdateQuestionParams) => {
```
boblauer: Rather than having `updateQuestion` update the state in place (and thus we need to pass in a copy of the state first), you could write it as a more traditional reducer that returns a copy of the state.

```js
function updateQuestion({ formOrQuestionGroup, action, updateFunction }) {
  return {
    ..formOrQuestionGroup,
    questions: formOrQuestionGroup.questions.map((questionOrQuestionGroup) => {
      if (isQuestionGroup(questionOrQuestionGroup)) {
        return updateQuestion({ formOrQuestionGroup: questionOrQuestionGroup, action, updateFunction });
      }

      return questionOrQuestionGroup.key === action.key ?
        updateFunction(question, value) :
        question
    });
  }
}
```

`updateFunction` would then also be able to return a full copy of the question, rather than updating the `question` in place.

mellogarrett: Tried the recursive approach, but had some issues with types. 

I found a pretty clean way to abstract this logic into a few steps. Let me know what you think.

PR #291, opened 2019-10-30
Reorder server startup function order

mellogarrett:
## Jira Issue

n/a

## Description

I've been having issues locally where projections are throwing errors saying they haven't been `connect()`ed before I called them. Somewhat counterintuitively, running `atlas.connect()` _after_ `app.prepare()` fixes that issue.

## Type of change


server/index.ts
```diff
@@ -15,19 +15,19 @@ async function main() {
...
-  await atlas.connect();
-  logger.info('connected to the database');
-
   const portArg = process.argv[process.argv.findIndex((arg) => arg === '-p') + 1];
   const port = parseInt(process.env.PORT || portArg || '8809', 10);
   const dev = process.env.NODE_ENV !== 'production';
   const app = next({ dev });
   const handle = app.getRequestHandler();
 
+  verifyEnvVars(['DATABASE_URL', 'LAMBDA_ENV', 'LOG_LEVEL', 'NODE_ENV']);
```
brentjanderson: Don't we want to `verifyEnvVars` before using `NODE_ENV` above? If we consider `NODE_ENV` required then it would seem to be necessary for our check above? I'd suggest moving `verifyEnvVars` and `logBasicEnvInfo` back up top, but the rest can go down below IMO.

mellogarrett: Good point

boblauer:
I'm nervous to merge this without understanding the underlying cause of why it was breaking before (only sometimes, so there's a race condition somewhere) and why this fixes it (If it's a race condition, I wouldn't expect re-ordering to fix it).

mellogarrett:
Yea it's weird. The error was present 100% of the time when I made a SQL edit in a local `atlas-db` projection, and referenced the local `atlas-db` code within Forms. 

I did find [the PR](https://github.com/LambdaSchoolEngineering/forms/commit/0383f4c2727284e8e525a3f9553b6275e96e8525#diff-9076f85cacae2708e2ec6e5b58e63e41) where this was reordered to what it is right now. This PR would actually just revert the order back to what we started with.

mellogarrett:
Unfortunately, the Next documentation is pretty sparse around custom servers and just points to a few examples. There is not mention of how `app.prepare()` works in detail.

Here's the example we originally used: https://github.com/zeit/next.js/blob/master/examples/custom-server/server.js

boblauer:
I'm fine with pushing through. Let's verify the staging heroku dyno launches correctly and test it out, and then assuming that works we'll just have to keep an eye on production.

mellogarrett:
Looks like `app.prepare()` only does something in [next-dev-server](https://github.com/zeit/next.js/blob/c2a2417e0c4d35111fffb46154adacd87ffe7b6f/packages/next/server/next-dev-server.ts#L195). In the production [next-server](https://github.com/zeit/next.js/blob/c2a2417e0c4d35111fffb46154adacd87ffe7b6f/packages/next/next-server/server/next-server.ts#L214), it's just a stubbed method.

So this likely won't have any impact on staging or production.

boblauer:
Thank you for digging!

mellogarrett:
I don't see any issues in staging at this point.

PR #318, opened 2019-11-12
Forms v2 discussion

mellogarrett:
I wanted to start the discussion around what Forms v2 should look like. I've started trying to imagine what this project would look like with a few things re-imagined, the biggest of all being how/when we fetch data.

Currently, we only fetch up front and we do that with `Projections`. The problem here is that these projections can return a lot of data, while still limiting access to the majority of the data in the database. They aren't very flexible to a user's needs. Take for example the Module Review form. The options for what Module Retro to review is a list of thousands of options that must be loaded into the browser. But at the same time, that list is limited to the past week of submissions. We are basically just moving a portion of the database into the browser and doing the querying there.

To make this process more flexible, I'm proposing we move to three new options for fetching data, and generally moving more logic to the server rather than the client. The first two options have to do with the initial loading of the form:

1) Prepopulate a list of options for a question with a scoped database query. Similar to what we have now, but implemented a bit simpler. Since we now own the server _and_ the client, we can remove a lot of logic from the client. I propose we remove the `prepopulateChoices` Question attribute and instead decide which questions get prepopulated (and how) in the Next API endpoints exclusively. 
2) Prepopulate a single option with a query parameter. Basically, if a query param exists for a given question key, then only query the database for the single record needed. If no query param exists, fallback to option 1 above where we query the database for a scoped list of options.

The third option for loading data would be initiated from the client _after_ the initial load and would be triggered by user interaction:

3) Live search, enabled by a new `searchUrl` Question property. This would only be available for `QuestionType.Search` and would require a search string from the user to execute a scoped database query.

I've scaffolded out the backend code for the Module Review form as an example of what this would look like. Let me know what you all think, if you have questions, concerns, or think I'm completely off-base here 😅


boblauer:
In general I'm a fan of this, and agree we need to start rethinking our approach to getting data from the database.

1. Agree, nothing to add here.

2. I go back and forth on this. Prepopulating a value but allowing the user to change it could be useful in some scenarios (I think). Either way we'd need to account for invalid values, in which case we'd need to fall back to treating it as if no query param were passed in at all.

3. I like the idea, but we'll have to be careful about the queries we write and the indexes needed to support them. Searching across all 100k unreviewed Module Submission records is going to take some careful consideration.

One idea would be to only search across Module Submissions that are from students in the same section as the TL filling out the form. In that case, we'd also want the ability to pass in the TL record they selected from the "Your Name" question. Perhaps all of the form's current values (or at least all dynamic dropdown values) are passed in the request body by default to the searchUrl endpoint, so that in there we can inspect their previous answers and act accordingly.

mellogarrett:
> Prepopulating a value but allowing the user to change it could be useful in some scenarios (I think)

I agree. I can envision it such that you can provide a query param to pre-populate a value on initial load, and also define a `searchUrl` on the question so that a user could clear the pre-populated value and choose a new value if they wanted.

mellogarrett:
> One idea would be to only search across Module Submissions that are from students in the same section as the TL filling out the form.

I believe we started by trying to do this and it broke the concept of TLs covering for one another.

I definitely agree this search would need to be thought through carefully. I haven't put a lot of thought into this piece.

boblauer:
> I believe we started by trying to do this and it broke the concept of TLs covering for one another.

I can't remember if that was the case. I know only showing students that are directly under the TL breaks that scenarios, but I'm not sure if TLs cover for each other across sections.

mellogarrett:
Oh right. I think what we actually had was showing students where `current_project_manager` = my TL record.

MarkDKoslow:
Yeah I'm in agreement on 1 and 2 as well. For 3, I think it'd be useful but I'm not sure it'd be super high priority to implement. Module submissions seem like an extreme case for loading data into the forms, and even that seems to be working relatively well for the time-being. 

Based on the [forms spec](https://docs.google.com/document/d/1ooYPUqRBxyn4bFaYkY1qaB7eE-kZI39U_DFSQSdE24o/edit?ts=5d672068#), it looks like conditionally showing/hiding questions based on user state is going to be coming up as well. I'm not sure if this draft is the place to start planning for that, but it looks like product/instruction would ideally have forms be dynamic based on a student's section (instruction, labs, X) and also their point in the curriculum (as for the career_upload questions). If so, there will be some architectural questions to address there as well.

MarkDKoslow:
> > I believe we started by trying to do this and it broke the concept of TLs covering for one another.
> 
> I can't remember if that was the case. I know only showing students that are directly under the TL breaks that scenarios, but I'm not sure if TLs cover for each other across sections.

We could also broaden this to courses instead of sections. I know some of the smaller cohorts in the past have only had one TL, so I would assume a TL from another section probably has to step in for them if they're out.

mellogarrett:
> For 3, I think it'd be useful but I'm not sure it'd be super high priority to implement.

Yea, you're probably right. It's probably the hardest piece to implement as well. 

> Based on the forms spec, it looks like conditionally showing/hiding questions based on user state is going to be coming up as well.

I've thought about this piece as well and I _think_ it should be pretty trivial to implement actually. We'd first need to assign `id`s to each Question. Then, at a basic level, the syntax could then look something like:
```ts
interface VisibleIf {
  questionId: string;
  answer: Answer
}

interface Question {
  ...
  visibleIf: VisibleIf;
}
```
Then we could use our existing `getAnswerForQuestion` util (and a new util called `getQuestionById`) to check if a question should be visible. Within `QuestionComponent`, something like:
```
let isVisible = true;

if (question.visibleIf) {
  const parentQuestion = getQuestionById(form, question.visibleIf.questionId);
  if (getAnswerForQuestion(parentQuestion) !== question.visibleIf.answer) {
    isVisible = false;
  }
}
```

boblauer:
Something I think about from time to time is "are you doing ourselves a disservice by having all of these forms driven by a config object?". Which is kind of another way of asking "will we ever save these configs in a database somewhere where they are dynamically generated/modified by people outside of eng?".

If the answer to that last question is no, then I think we're hurting ourselves by having these be config driven. We're not too hampered currently, but as we start adding new features, such as hiding and showing fields conditionally, fitting all of that logic into a config becomes difficult. We need to start writing our own language that we can save as strings and then convert that into executable javascript (since we can't serialize javascript functions themselves).

You could imagine any of these forms being a list of different question components within the form file itself. By doing that, it would allow us to write one-off onChange or show/hide logic easily. Adding question sections (already done but wasn't trivial) does become trivial. I would argue that everything becomes easier if we treat the form as a normal React page.

However, if we go down this route, reversibility is low.

However however, if we keep down the config-driven route in anticipation for a day  when we store these configs in a database and that day never comes, we've made our lives much more difficult and hampered the speed that we can start implementing more complex features.

Part of me thinks that database-driven forms will never happen, and we're hurting ourselves in the present for a possible future that won't happen. But another part of me thinks it _might_ come, and it would be difficult to add onto a non-config driven approach.

Generally speaking I'm a fan of [YAGNI](https://www.martinfowler.com/bliki/Yagni.html), and I've seen examples of code made way more complicated then it needed to be in preparation for a future that never came.

This could be a good topic for our Forms Weekly meeting in 5 minutes.

mellogarrett:
The possibility of storing these in a database was not a driving factor in making these forms config-driven. The config provides a lot of good value that would be hard to imitate without a config. 
- Consistent UI & UX across different forms (brand continuity)
- One file to look at for Form definition (declarative)
- Good type safety
- Less mental load when switching between Forms because of consistencies; You know what to expect
- Shared store logic across all forms

I fear without a config we'd wind up with a lot of one-off styles, data fetching implementations, state handlers, etc. that would quickly make the codebase miserable to work in.

boblauer:
> Consistent UI & UX across different forms (brand continuity)

I think we still get this by using shared components.

> One file to look at for Form definition (declarative)

I think we still would have one file to look at for any given form, but it would be the `pages/some-form.ts` file rather than `config/some-form.ts`

> Good type safety

Using the same shared components should give us type safety through `props`

> Less mental load when switching between Forms because of consistencies; You know what to expect

I wouldn't expect one form's code to the next to look so wildly different than it would be difficult to switch between them. I think we'd still be shooting for code consistency

> Shared store logic across all forms

I would assume we could still implement this even if all components were listed explicitly in any given form.

> I fear without a config we'd wind up with a lot of one-off styles, data fetching implementations, state handlers, etc. that would quickly make the codebase miserable to work in.

I suppose this is possible, although this seems possible for the vast majority of React apps since most are not driven by a config. In any React app I've worked on, there was usually an established way of fetching data, managing state, etc. I never found those things to get out of hand.

I think the config approach has generally served us well so far, but I also think that's because we haven't started implementing more complicated features. And I also know that we _can_ implement nearly any feature using a config-driven approach, but I also think for any non-trivial feature, implementing it in a config-driven app is going to be far more difficult than in a standard React app.

I'm not suggesting we scrap the config-driven approach right now, but I do think we should keep our eyes open for when we find ourselves bending over backwards to make a feature work and the only reason it's going to take 2 weeks instead of 2 hours is because of the config. (This was kinda true for Section headings, although it ended up taking 1-2 days instead of a few hours and we were fine with that).

MarkDKoslow:
It may be helpful to chat with Aparna & Caleb about the future vision for Forms to get a better sense for how much more technically complex these are going to get over the coming months/years. Obviously they won't have all the answers, but at least we'll have as much information as possible, based on the current vision for the app. I also wonder if there's some middle ground solution here, where the form questions are offloaded into config files but any custom show/hide logic, section headers, UX concerns, etc. are included in form-specific layout components.

brentjanderson:
A key consideration for much of this is if we ever expect our users to be able to craft their own forms. If being self-service is ever on the roadmap, then this becomes a problem. As long as engineering support is required for reading/writing the data (and I expect that is likely to be the case), then we will probably not support self-service forms, reducing that as a constraint on what we build in the fturue.

mellogarrett:
I personally don’t see us ever getting to a place where users can build their own Forms. We might be able to create a tool that can spit out our config  object, but that’s just the top of the iceberg. The majority of our work and the more complex pieces are creating API endpoints and custom context methods bespoke to our use-cases.

brentjanderson:
Agreed about the API endpoints. If we did want to go that route, JSON Schema would be something to consider, as it can be used to render forms: https://rjsf-team.github.io/react-jsonschema-form/

That said, I agree that we shouldn't shy away from writing standard React and calling it a day.

mellogarrett:
I don’t think we need to involve a ton of people at this point. The fundamental features for v2 have already been defined. Just trying to think through the technical details of how to implement the features already discussed.

mellogarrett:
Bob, sounds like that implementation wouldn’t be as different from a config-driven approach as I imagined then.

mellogarrett:
Given the feedback, I think we can punt on a lot of this. Regarding option #1, I think we can greatly simplify/delete a lot of code around _how_ we do this currently without changing _what_ we are doing. We might want to forward query params to the backend and try to only fetch a single object by id, but even that is a big maybe. Like Bob mentioned, you still might want to clear that pre-selected answer and choose from the full list anyway. In regards to live search, it's probably less useful that I first imagined and just presents a bunch of new questions. 

I think the big-ticket v2 items (query param support and conditional question chains) are completely doable with our current approaches to these problems. 

I'd also be open to seeing what a config-less approach to one of these forms would look and feel like. I'll try to be cognizant of when the config starts getting in the way, and maybe at that point someone can put up a config-less PR to get feedback.

MarkDKoslow:
That sounds good to me. Thanks for driving this conversation and scaffolding out the sample code @mellogarrett

PR #323, opened 2019-11-15
[CORE-909] Auto-select option if suggestedValue is true

MarkDKoslow:
## Jira Issue
[CORE-909](https://lambdaschool.atlassian.net/browse/CORE-909)

## Description
Auto-select option if suggestedValue field is true and another choice has not already been selected

## Type of change


http/merge-question-options/forms-api/transform-grouped-forms-api-options/index.ts
```diff
@@ -6,16 +6,23 @@ export const transformGroupedFormsApiOptions: GroupedOptionsToChoiceArrayTransfo
...
         display,
-        checked: false,
+        checked: selected && !selectedChoiceForQuestion ? true : false,
         detail,
       };
 
+      // Record that an option has been selected, so no further options can be selected for that group
+      if (choice.checked && !selectedChoiceForQuestion) {
+        selectedChoiceForQuestion = true;
+      }
```
mellogarrett: Hmm seems weird that we have to do this. Seems like the backend should only return one `suggestedValue`?

MarkDKoslow: Yeah i was going back and forth on this as well. I think it'd probably be a better UX in some cases to return a few suggestedValues, like if we can narrow their probable instructors to 2 or 3 options, but aren't sure exactly which is correct. In that case, we'd choose 1 randomly but the other 2 would be the next two options in the dropdown. This would also be a fallback in case we forget to limit suggested_value query to 1 result. However, I could also see the case that we'd want to enforce only having 1 suggested_value per question.

MarkDKoslow:
@mellogarrett I'm going to hold off on merging till Monday but lmk if you have any other thoughts on this

PR #325, opened 2019-11-18
Add conditional question support

mellogarrett:
## Jira Issue

[CORE-1323]

## Description

- Add `visibleIf` property where questions can define when they should be visible.

**There are lots of files changed, mostly because this required adding an `id` to each question.**

## Type of change

[CORE-1323]: https://lambdaschool.atlassian.net/browse/CORE-1323


types/questions.ts
```diff
@@ -27,12 +28,18 @@ export interface Range {
   end: number;
 }
 
+interface VisibleIf {
+  questionId: string;
+  answer: Answer;
+}
+
 interface BaseQuestion {
+  id: string;
```
boblauer: Any reason we can't use `key` for `visibleIf` instead of adding an `id`?

mellogarrett: Oooooh we can totally use `key`. Brain fart there.

mellogarrett:
Okay, this is updated to use `Question.key` instead of a new `Question.id`

utils/is-question-visible/index.ts
```diff
@@ -0,0 +1,17 @@
+import { Question, Form } from '../../types';
+import { getQuestionByKey } from '../get-question-by-key';
+import { getAnswerForQuestion } from '../get-answer-getter-for-question-type';
+
+export function isQuestionVisible(question: Question, form: Form) {
+  if (question.visibleIf) {
```
boblauer: This can be simplified.

```ts
if (!question.visibleIf) return true;

const parentQuestion = getQuestionByKey(form, question.visibleIf.questionId);
return parentQuestion && getAnswerForQuestion(parentQuestion) === question.visibleIf.answer;
```

utils/is-question-visible/index.test.ts
```diff
@@ -0,0 +1,68 @@
...
+      key: 'parent' as any,
+    };
+    const question: Question = {
+      type: QuestionType.LongText,
+      title: "I'm conditionally visible",
+      isRequired: true,
+      key: '' as any,
+      visibleIf: {
+        questionId: parentQuestion.key,
+        answer: ['A', 'B', 'C'],
```
boblauer: In the case of a multi-select dropdown, even if A B and C were all checked above, this question would still not be visible because `question.answer` is going to be a different array than `getAnswerForQuestion(parentQuestion)`, so the equality check will always be false.

If the answer we get back is an array, we probably want to `join` it into a string, and then `join` our `visibleIf.answer` into a string as well, and compare those two string. Worth adding a test for that as well.

mellogarrett: Ah nice catch. I think we'd also want to consider `['A', 'B', 'C']` to be the same as `['A', 'C', 'B']` for this check as well, right?

boblauer: Yes, good call, we should `sort` before `join`ing.

mellogarrett: I'm going to go ahead and use lodash `isEqual` since that's already a dep.

utils/derive-form-submission/index.test.ts
```diff
@@ -0,0 +1,22 @@
...
+        visibleIf: {
+          questionId: 'test',
+          answer: 'doesnt exist',
+        },
+        key: 'test' as any,
+      },
+    ]);
+
+    expect(deriveFormSubmission(form)).toEqual({});
+  });
```
MarkDKoslow: It may be worth adding a few more test cases here. There could be one for questions and one for question groups as well. I'd also just want to make sure `addAnswerToFormSubmission ` is working properly for visible questions

utils/get-question-by-key/index.ts
```diff
@@ -0,0 +1,26 @@
+import { Form, isQuestionGroup, Question } from '../../types';
+
+export function getQuestionByKey(form: Form, key: string) {
+  let question: Question | undefined;
+
```
MarkDKoslow: Could you flatten the questions array instead and then search through it? Something like 
`
const questions = [].concat(...form.questions);
return questions.find((question) => question.key === key)
`

MarkDKoslow: Oh I see, QuestionGroup is an object not an array

MarkDKoslow: We could do something like:
```const flattenedQuestions = forms.questions.map((question) => isQuestionGroup(question) ? question.questions : question).flat();```
Looks like we'd need to add es2019 to our --lib settings to get `flat` though https://stackoverflow.com/questions/53556409/typescript-flatmap-flat-flatten-doesnt-exist-on-type-any

Comment (mellogarrett) [reply utils/get-question-by-key/index.ts]: I think we could achieve a similar thing with `.reduce` without needing the es2019 syntax. I'll look into this.

File: utils/is-question-visible/index.test.ts (hunk 3 above)
Comment (mellogarrett) [reply utils/is-question-visible/index.test.ts]: should be good now

File: utils/get-question-by-key/index.ts (hunk 5 above)
Comment (mellogarrett) [reply utils/get-question-by-key/index.ts]: *much* better now

File: utils/validate-form/index.ts [hunk 6]
```diff
@@ -1,8 +1,9 @@
 import isEqual from 'lodash.isequal';
 import { Form, Question, isQuestionGroup, Choice } from '../../types';
+import { isQuestionVisible } from '../is-question-visible';
 
 export const validateForm = (form: Form): boolean =>
-  areAllQuestionsPassingValidation(form) && areAllRequiredQuestionsAnswered(form);
+  areAllQuestionsPassingValidation(form) && areAllVisibleRequiredQuestionsAnswered(form);
```
MarkDKoslow: I'm thinking we should also update `areAllQuestionsPassingValidation` to skip non-visible questions. Otherwise, if you had a validation error on that field, and then the field was hidden, the user wouldn't be able to submit but wouldn't be able to see why

mellogarrett: True, good catch

test/forms.test.ts
```diff
@@ -0,0 +1,27 @@
...
+describe('All Forms', () => {
+  test('Each Form should have unique Question/QuestionGroup keys', () => {
+    Object.entries(Forms).forEach(([, form]) => {
+      const ids = form.questions.reduce((idArr: string[], questionOrQuestionGroup) => {
+        if (isQuestionGroup(questionOrQuestionGroup)) {
+          return [...idArr, ...questionOrQuestionGroup.questions.map((question) => question.key)];
+        }
+        return [...idArr, questionOrQuestionGroup.key];
+      }, []);
+
```
MarkDKoslow: Because we just want to know each id is unique, I think you could also just pass the ids array into a set and compare the length here. Something like:
```
var arr = [55, 44, 65];
var set = new Set(arr);
console.log(set.size === arr.length);
```

utils/derive-form-submission/index.test.ts
```diff
@@ -0,0 +1,111 @@
...
+        key: 'd' as any,
+        choices: [
+          { key: 'a', value: 'A', checked: true },
+          { key: 'b', value: 'B', checked: false },
+          { key: 'c', value: 'C', checked: true },
+        ],
+      },
+    ]);
+
+    expect(deriveFormSubmission(form)).toEqual({ a: null, b: 2, c: null, d: ['A', 'C'] });
```
MarkDKoslow: Thanks for adding these 👏

test/forms.test.ts
mellogarrett: The only downside with that approach is that it won't show you the offending key, just that _some_ key is duplicated.

mellogarrett:
Just added support for using `visibleIf` on `QuestionGroup`s as well. and updated `VisibleIf.questionId` to be `VisibleIf.questionKey`

mellogarrett:
Adding `visibleIf` to `QuestionGroup`s complicated the logic and created a bug in `isQuestionOrQuestionGroupVisible`, but I fixed it and that util's test file should be very thorough now.

mellogarrett:
^ @boblauer @MarkDKoslow

PR #332, opened 2019-11-25
Conditional posting

mellogarrett:
## Jira Issue

[CORE-1346]

## Description

- Add logic to Sprint Retro API endpoint to conditionally insert build week data.
- Add build week questions to Sprint Retro form.

**NOTE: requires the appropriate `atlas-db` context methods to be shipped first.**

## Type of change

[CORE-1346]: https://lambdaschool.atlassian.net/browse/CORE-1346


mellogarrett:
This is ready to go for release day

pages/api/sprint-retrospective.ts
```diff
@@ -21,8 +22,46 @@ export default async (req: NextApiRequest, res: NextApiResponse) => {
...
+    disruptiveTeamMembers: data.disruptiveTeamMembers,
+  };
+
+  const sprintSubmissionId = (
+    (await atlas.instruction.sprintSubmission.submitForm(userId, sprintSubmission)) || { id: undefined }
+  ).id;
+  if (!sprintSubmissionId) throw new Error('Error submitting this form');
+
+  if (isInBuildWeek) {
+    await atlas.instruction.sprintSubmission.addBuildWeekDataForSprintSubmission(
```
boblauer: This doesn't have to happen in this pull request and block going live with these forms, but this should all be done in a single context method so that it can be done in a single SQL transaction.

Generally speaking, if request B relies on the result of request A, they should be bundled within a single context method, and within that method wrapped in a single SQL transaction.

mellogarrett: Yea makes sense.

forms/sprint-retrospective.ts
```diff
@@ -207,5 +207,55 @@ Please use the following criteria:
...
+        {
+          title: 'Please select any of your team members who have gone above and beyond this Sprint',
+          description:
+            'This feedback will remain private, and only used to ensure each team member gets the support and coaching they need',
+          type: QuestionType.Dropdown,
+          isMultipleSelect: true,
+          prepopulateChoices: true,
+          choices: [],
+          isRequired: false,
+          key: 'aboveTeamMembers',
```
MarkDKoslow: Right now the projection is only returning one key for these three questions (`peers`). I remember thinking about this a few months ago, but do we have a way to have different keys for get/post?

boblauer: Ah, interesting. So we want to use the same response data for all three of these questions, but obviously want to submit the answers as different keys.

A few options come to mind:

---

- `key: { get: 'peers', post: 'aboveTeamMembers' }`

Pros: Clearly spelled out
Cons: Have to check if `key` is an object or string everywhere we expect it to be a string.

---

- `key: 'peers|aboveTeamMembers'` or `key: 'peers,aboveTeamMembers'`, etc.

Pros: Can use the key as-is in places where we're not concerned about the choices data
Cons: Not as obvious what's happening

---

- Change the response object that contains the choices to have the key `'aboveTeamMembers,effectiveTeamMembers,disruptiveTeamMembers'` and update the code that populates the initial `choices` to split on `,` and populate multiple questions.

Pros: This is probably the path of least resistance
Cons: Honestly not sure. This actually feels pretty clean, all things considered.

pages/api/sprint-retrospective.ts
```diff
@@ -21,11 +22,61 @@ export default async (req: NextApiRequest, res: NextApiResponse) => {
...
+    if (questionId === 'peerId') {
+      newFormData.push({ questionId: 'aboveTeamMembers', ...rest });
+      newFormData.push({ questionId: 'effectiveTeamMembers', ...rest });
+      newFormData.push({ questionId: 'disruptiveTeamMembers', ...rest });
+    }
+
+    newFormData.push({ questionId, ...rest });
+
+    return newFormData;
+  }, []);
```
mellogarrett: @boblauer @MarkDKoslow thoughts on this approach? I'm just cloning the `peerId` options for each unique post key.

boblauer: Only downside is that we're sending 4x the amount of data we need to back to the browser (the original `peerId` object, and 3x clones of it).

I would be inclined to do this step on the client rather than the server, especially for this scenario where we're returning quite a few options.

forms/sprint-retrospective.ts
mellogarrett: Since we use `Question.Key` in many places, changing it would be pretty invasive. I opted to just add an optional `Question.prepopulateKey`. Let me know what you think.

PR #350, opened 2019-12-16
POC of a config-less form

mellogarrett:
## Jira Issue

n/a

## Description

This is my attempt at converting the Forms app to be a bit more flexible by changing the starting point when creating new forms from a config file, to a React component. The example implementation is [here](https://github.com/LambdaSchoolEngineering/forms/blob/look-ma-no-config/pages/no-config.tsx).

The only Question Component I've converted so far is the `CheckboxQuestionComponent`. `visibleIf` is now implemented as a prop on the question components called `isVisible: boolean`. `SubmitButton` now takes a single URL to post to, doing away with all of the `Backends[]` complexity. There is a new `useManagedQuestion` hook that handles adding/replacing questions in the store and toggling their visibility.

If we eventually convert all forms to this approach, we'll be able to delete a substantial amount of code from the project.

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)


http/submit-form/index.ts
```diff
@@ -1,29 +1,13 @@
...
-      })
-      .catch((error) => {
-        window.logger.error(new LError(`Error posting form data to ${adapter.label}.`, error));
-        throw error;
-      })
-  );
-
-  return submitFlow(submitActions);
+  const atlas = await createAtlasClient(Axios);
+  atlas.post(postUrl, JSON.stringify(deriveFormSubmission(form)));
```
MarkDKoslow: 🙌

pages/no-config.tsx
```diff
@@ -0,0 +1,71 @@
...
+        form={form}
+        dispatch={dispatch}
+      />
+      <CheckboxQuestionComponent
+        question={{
+          title: 'Conditional',
+          isRequired: true,
+          key: 'child' as any,
+          choices: childChoices,
+          isVisible: isEqual(getAnswerForQuestion(getQuestionByKey(form, 'dropdown')), 'show child'),
```
MarkDKoslow: For resubmissions, I think we'll just need to store some metadata on the section sprint question about whether they chose a resubmission, and then can call something like `getMetadataForQuestion`

store/reducers/add-question.ts
```diff
@@ -0,0 +1,7 @@
+import { Reducer } from 'react';
+import { Form, AddQuestionAction } from '../../types';
+
+export const addQuestionReducer: Reducer<Form, AddQuestionAction> = (prevState, action): Form => ({
+  ...prevState,
+  questions: [...prevState.questions, action.value],
```
MarkDKoslow: Would the questions here have to be added to the end of the questions array? I'm thinking in the case of BW & Unit submissions, where, if they select "Yes" for "Is this a BW?", we'd want to add questions before the "Is this a unit submission?" question

pages/no-config.tsx
mellogarrett: Yep, for sure. Or something like 
```ts
isVisible: selectedSectionSprint.isBuildWeek === true
```

It'd be super easy to do and we wouldn't have to account for `isBuildWeek` in the entire Forms app codebase, just in the form it applies to.

store/reducers/add-question.ts
mellogarrett: Since we updated all the redux stuff to use `question.key` rather than question index, I don't think order really matters. I think in the Forms API endpoint, we'd just pull out the different sets of answers by key similar to what we're already doing [here](https://github.com/LambdaSchoolEngineering/forms/blob/43f1217aab7b0b68e5bc16640e5084f10699423b/pages/api/sprint-retrospective.ts#L27-L50).

PR #357, opened 2019-12-20
Choice metadata

mellogarrett:
## Jira Issue

n/a

## Description

This PR adds the functionality to conditionally show questions or question groups based on another question's answer metadata.

In practice, this will allow us to get rid of the "Are you in a build week?" question and instead key off of the Section Sprint metadata to determine this automatically. So then when they answer "What did you study during this sprint?" the Build Week questions' visibility will automatically be toggled based on their answer.

## Type of change


mellogarrett:
There is some noise in the file changes because I moved around some of the utils.

PR #360, opened 2020-01-02
Add Unit Review questions to Sprint Review form

mellogarrett:
## Jira Issue

[CORE-1605]

## Description

- Conditionally add the Unit Review questions to the end of the Sprint Review form based on if the TL selects a Sprint Submission that has a corresponding Unit Submission or not.
- Update Sprint Retro & Review API code to the updated `atlas-db` interface.
- Remove standalone Unit Retro & Review forms.
- Update `question-key` types to accommodate iterating over the key sets.

## Type of change

[CORE-1605]: https://lambdaschool.atlassian.net/browse/CORE-1605


mellogarrett:
This build is failing because it needs the new atlas-db version.

forms/sprint-review.ts
```diff
@@ -133,5 +133,91 @@ Student can access this feedback through their dashboard`,
...
+          range: { start: 1, end: 3 },
+          title: 'Unit Assessment Score 6',
+          isRequired: false,
+          key: 'unitRubricItem6Rating',
+        },
+        {
+          type: QuestionType.Number,
+          title: 'On a scale of 1-10, how likely would you be to hire this student?',
+          isRequired: true,
+          key: 'hireabilityNpsRating',
```
MarkDKoslow: Can we still restrict these to a range? I think we do for NPS

forms/sprint-review.ts
```diff
@@ -133,5 +133,91 @@ Student can access this feedback through their dashboard`,
...
+        {
+          type: QuestionType.Rating,
+          range: { start: 1, end: 3 },
+          title: 'How did the student demonstrate mastery of the unit content?',
+          isRequired: true,
+          key: 'unitOverallRating',
+        },
+        {
+          type: QuestionType.LongText,
+          title: 'What is blocking the student from passing?',
```
MarkDKoslow: More of a product question but this seems like weird wording cause it assumes they didn't pass 🤔. Maybe something like "If the student has not passed the unit, what is blocking them from doing so?" Or we could hide this question unless the previous question is set to a 1

forms/sprint-review.ts
mellogarrett: Yea good call

forms/sprint-review.ts
mellogarrett: This is what was on the spec, but I agree it's not worded well. That's a great idea to make it conditional on the question above. I'll reach out to Aparna & Mitchell on this.

pages/api/sprint-review.ts
```diff
@@ -21,8 +22,40 @@ export default async (req: NextApiRequest, res: NextApiResponse) => {
...
+    unitTechnicalRating: data.unitTechnicalRating,
+    unitTeamworkRating: data.unitTeamworkRating,
+    unitRubricItem1Rating: data.unitRubricItem1Rating,
+    unitRubricItem2Rating: data.unitRubricItem2Rating,
+    unitRubricItem3Rating: data.unitRubricItem3Rating,
+    unitRubricItem4Rating: data.unitRubricItem4Rating,
+    unitRubricItem5Rating: data.unitRubricItem5Rating,
+    unitRubricItem6Rating: data.unitRubricItem6Rating,
+    hireabilityNpsRating: data.hireabilityNpsRating,
+  };
```
MarkDKoslow: Hmm it'd be cool if there were a shorthand way to extract all the keys/values from an object that conform to a type (i.e. the `Keys` types above). I think we could in a loop but not sure that gives us much

mellogarrett: Yea agreed. I'll try to figure this one out.

boblauer: You can get pretty close:

```ts
// types/question-keys.ts

export const unitSubmissionReviewProps = [
  'unitOverallRating',
  'blocker',
  'unitTechnicalRating',
  'unitTeamworkRating',
  'unitRubricItem1Rating',
  'unitRubricItem2Rating',
  'unitRubricItem3Rating',
  'unitRubricItem4Rating',
  'unitRubricItem5Rating',
  'unitRubricItem6Rating',
  'hireabilityNpsRating',
] as const;

export type UnitSubmissionReviewKeys = typeof unitSubmissionReviewProps[number];

// api/sprint-review.ts

import { unitSubmissionReviewProps, UnitSubmissionReviewKeys } from '../../types';

async function submitForm(...) {
  const unitReviewSubmission = _.pick(data, unitSubmissionReviewProps);
}
```

Only gotcha is that the `types` file now contains actual javascript and not just types anymore, but being able to define a type and use the array of props from a single array seems like that's worth the trade off.

An alternative would be to define the list of keys in an array in the `api/sprint-review.ts` file, and then define the type within that same file.

```ts
const unitSubmissionReviewKeys = ['a', 'b', 'c'] as const;

async function submitForm(userId: string, data: Record<typeof unitSubmissionReviewKeys>) {
  const unitReviewSubmission = _.pick(data, sprintSubmissionReviewKeys);
}
```

pages/api/sprint-review.ts
```diff
@@ -21,8 +22,40 @@ export default async (req: NextApiRequest, res: NextApiResponse) => {
   }
 };
 
-async function submitForm(userId: string, data: object) {
-  await atlas.instruction.sprintSubmissionReview.submitForm(userId, data);
+async function submitForm(userId: string, data: Record<SprintSubmissionReviewKeys | UnitSubmissionReviewKeys, Answer>) {
```
boblauer: One thing that stands out here is the types are `XSubmissionReview`, but the data objects are `xReviewSubmission`. Is the transposing of Submissions and Review on purpose?

mellogarrett: Nope, not on purpose. I'll clean these names up a bit.

pages/api/sprint-review.ts
mellogarrett: Can we use an enum instead of a `type` of pre-defined strings? enums are iterable and live on at runtime.

mellogarrett: Like 
```ts
export type SprintReviewKeys =
  | 'careerRating'
  | 'careerFeedback'
  | 'interventionReferral'
  | 'otherNotes'
  | 'otherInternal'
  | 'projectManagerId'
  | 'reviewEscalation'
  | 'skillsFeedback'
  | 'skillsRating'
  | 'sprintSubmissionId'
  | 'tags'
  | 'teamworkFeedback'
  | 'teamworkRating'
  | 'threeWordSummary'
  | 'whatWentPoorly'
  | 'whatWentWell';
```
becomes
```ts
export enum SprintReviewKeys {
  careerRating = 'careerRating',
  careerFeedback = 'careerFeedback',
  interventionReferral = 'interventionReferral',
  otherNotes = 'otherNotes',
  otherInternal = 'otherInternal',
  projectManagerId = 'projectManagerId',
  reviewEscalation = 'reviewEscalation',
  skillsFeedback = 'skillsFeedback',
  skillsRating = 'skillsRating',
  sprintSubmissionId = 'sprintSubmissionId',
  tags = 'tags',
  teamworkFeedback = 'teamworkFeedback',
  teamworkRating = 'teamworkRating',
  threeWordSummary = 'threeWordSummary',
  whatWentPoorly = 'whatWentPoorly',
  whatWentWell = 'whatWentWell',
}
```

mellogarrett: Wow, that `as const` is pretty neat: https://blog.logrocket.com/const-assertions-are-the-killer-new-typescript-feature-b73451f35802/

And I guess the `[number]` just means any index of that const?

boblauer: Yup, exactly. Got it from here - https://dev.to/shakyshane/2-ways-to-create-a-union-from-an-array-in-typescript-1kd6

PR #366, opened 2020-01-07
Comma Separated URLs

mellogarrett:
## Jira Issue

n/a

## Description

Create a validation function for a list of comma-separated URLs and apply that validation to the `buildWeekAdditionalAssessment` question.

## Type of change


constants/validations/url/index.ts
```diff
@@ -7,5 +7,14 @@ const isValidUrl = (input: string): boolean => urlValidationRegex.test(input);
...
-  error: 'Please enter a valid url',
+  error: 'Please enter a valid URL',
+};
+
+export const commaSeparatedUrlsValidation: Validation = {
+  validate: (input: string): boolean =>
+    input
+      .split(',')
+      .map((chunk) => chunk.trim())
+      .every((chunk) => isValidUrl(chunk)),
```
MarkDKoslow: Not a blocker but it may also we worth removing any newline characters, I could see people entering a URL, and then hitting return and pasting the next one

mellogarrett: Yea good point.

mellogarrett: Right now newlines will fail validation. Should we allow them and make sure we parse them out correctly?

boblauer: If we're rendering a `<textarea>`, I think we should support new lines.

PR #381, opened 2020-01-15
[CORE-1766] Conditionally show/hide unit questions based on form metadata

MarkDKoslow:
## Jira Issue
[CORE-1766]

## Description
Issue with last PR was that I needed to relax the isEqual check to isMatch, because the form metadata is now an object with two keys "showBuildWeekQuestions" and "showUnitQuestions"

## Type of change

[CORE-1766]: https://lambdaschool.atlassian.net/browse/CORE-1766


utils/is-question-or-question-group-visible/index.ts
```diff
@@ -54,7 +55,7 @@ function checkQuestionOrQuestionGroupVisibility(questionOrQuestionGroup: Questio
       );
     }
 
-    return isEqual(parentMetadata, questionOrQuestionGroup.visibleIf.answer);
+    return !!parentMetadata && isMatch(parentMetadata, questionOrQuestionGroup.visibleIf.answer);
```
mellogarrett: Do you need `!!parentMetadata`? If you have to have something there, will just `return parentMetadata &&...` work?

MarkDKoslow: I think that would also work. It looks like all the other code paths for this function return a boolean, so I mostly added that to keep it consistent. I also just prefer it stylistically, but don't have a strong preference here

mellogarrett: Do you prefer it over the more explicit `parentMetadata !== undefined`? That would be more clear IMO. 

Doesn't matter much either way, just being picky :)

MarkDKoslow: Good call, I like `parentMetadata !== undefined` more. Updating 👍

PR #383, opened 2020-01-17
Refresh JWT on load

mellogarrett:
## Jira Issue

[CORE-1762]

## Description

Refresh JWT on Form load.

## Type of change

[CORE-1762]: https://lambdaschool.atlassian.net/browse/CORE-1762


http/merge-question-options/forms-api/index.ts
```diff
@@ -33,6 +33,12 @@ export const mergeQuestionOptionsFromFormsApi = async (form: Form, ctx: NextPage
   const cookie = ctx.req.headers.cookie;
   const axiosClient = createAxiosClient(cookie);
 
+  try {
+    axiosClient.get(`https://auth.${getDomain()}/refresh`);
```
josh-egan: I think you'll want to `await` this call. Or, if it's intended to be fire-and-forget, then you'll want to remove the `try/catch` block and instead use `axiosClient.get(/* url */).catch(err => /* log */)`

boblauer: Yeah, this is fire and forget, but as Josh alluded to, a `try` block won't catch an rejected promise unless it's `await`ed.

I think a `.catch(...)` is appropriate.

boblauer: This doesn't feel like the right place for this to live. If we didn't have any dynamic dropdowns on a form, this code would never run for that form, but the timeout issue would still be a problem.

I would probably run this in a `useEffect` inside of `FormComponent`

PR #387, opened 2020-01-24
Merge pull request #387 from LambdaSchoolEngineering/one-thru-five-peer-feedback

boblauer:
## Jira Issue

N/A

## Description

Allow 1 through 5 ratings (work has already been done in atlas-db to allow this). Also updated the description to allow for embedding html, and update the question descriptions with wording from Joshua.

## Type of change


boblauer:
Correct, it would just be a visual check.

forms/peer-review.ts
```diff
@@ -77,17 +138,51 @@ export const PeerReviewForm: Form = {
...
+        <ul>
+          <li>Does not do a fair share of the team’s work. Delivers sloppy or incomplete work.</li>
+          <li>Misses deadlines, is late, unprepared, or absent for meetings.</li>
+          <li>Does not assist other teammates.</li>
+          <li>Quits if work becomes hard.</li>
+        </ul>
+      </li>
+
+      <li>
+        <p><b>Did Not Meet Expectations</b>: Demonstrates behaviors described in both 1 and 3.
```
MarkDKoslow: It may just be because I'm not seeing it visually, but I don't understand what `1 and 3` mean here. Is it the bullets from above? And if so, what does it mean below for `3 and 5`?

boblauer: Yeah, it's confusing because this is an ordered list, meaning it's numbered, but those numbers don't show up in code. It makes more sense visually:

image

PR #388, opened 2020-01-24
Roll back atlas-db to pre v17

boblauer:
## Jira Issue

N/A

## Description

@MarkDKoslow mentioned yesterday that he wasn't comfortable going forward with the new section sprint projection changes until next week. However, those changes are already in atlas-db as v17, and my changes to support peer review answers of 1 through 5 were added on top of his changes.

So, to include my changes without his, I checked atlas-db out to the last commit before v17, updated the Joi schema to account for 1 through 5, and then published it as 16.7.1. So this one-off patch now contains my fix, but doesn't contain Mark's new projection.

## Type of change


boblauer:
On second thought, this may not work. I might just roll Mark's change back in atlas-db, deploy that as a normal version, and then use that.

PR #414, opened 2020-02-05
[CORE-1699] Sprint resubmissions

MarkDKoslow:
## Jira Issue
[CORE-1699](https://lambdaschool.atlassian.net/browse/CORE-1699)

## Description
Update to the forms API to conditionally post to submitForm or submitResubmissionForm methods based on whether the instructor_id field has been filled out. This field seemed the least likely to be removed in the future.

This is admittedly brittle. However, Garrett and I spent a few hours working through other approaches this morning, and determined that this is more reasonable, at least until we can move to a config-less form world.

One other thing to note is that many questions will now be hidden by default, and will appear on the page (below the fold) if they choose a section sprint that they're not resubmitting for. If we want to change this, we can add a `hiddenIf` field, but I recommend we do it in a follow-up PR.

Test Cases
1. Normal Sprint Retro ✅ 
2. Resubmission Sprint Retro ✅ 
3. Normal Build Week Retro (Not tested yet)
4. Resubmission Build Week Retro (Not tested yet)
5. Normal Labs Retro (Not tested yet)
6. Resubmission Labs Retro (Not tested yet)

## Type of change


pages/api/sprint-retrospective.ts
```diff
@@ -56,12 +56,22 @@ async function submitForm(
...
   }
 
-  await atlas.instruction.sprintSubmission.submitForm(
-    userId,
-    sprintRetrospective,
-    Object.keys(sprintRetrospectiveBuildWeek).length > 0 ? sprintRetrospectiveBuildWeek : undefined,
-    Object.keys(unitRetrospective).length > 0 ? unitRetrospective : undefined
-  );
+  // If the student filled out the instructor field, then they're not resubmitting
+  if (sprintRetrospective.instructorId) {
```
mellogarrett: I would recommend you instead create a function to assert the type of the `sprintRetrospective` object. Something like:
```ts
interface SprintResubmission {
  ...
}

export function isSprintResubmission(x: {}): x is SprintResubmission {
  // assert that only the five(?) required fields for a resubmission are on the object
  return 'field1' in x && 'field2' in x && ... && Object.keys(x).length(5);
}
```

Then you can say `if (isSprintResubmission(sprintRetrospective)) ...`

boblauer:
Haven't looked at the code, but from a high level, keying off of `instructorId` is a clever solution and the right one. Also have no problem with questions showing/hiding based off of the section sprint the student chooses.

pages/api/sprint-retrospective.ts
MarkDKoslow: Hm I have mixed feelings here. I agree that having a `isSprintResubmission` function is more clear than just keying off of instructorId. However, by checking all 5 keys, I think we'd be expanding the potential for a future change to break something.

mellogarrett: True.

PR #417, opened 2020-02-06
Fix AtlasDomain export

mellogarrett:
## Jira Issue

n/a

## Description

Export an AtlasDomain getter rather than the static object so we can catch errors during its creation.

## Type of change


mellogarrett:
Going to go ahead and update to `atlas-db@21.2.0` as well, which should fix the underlying error here.

PR #423, opened 2020-02-11
Add no-config CheckboxQuestionComponent

mellogarrett:
## Jira Issue

n/a

## Description

- Create "no-config" version of the `<CheckboxQuestionComponent/>`
- In "no-config" `<FormsComponent/>`, add margin between questions

## Type of change


no-config-components/dropdown-question/index.tsx
```diff
@@ -19,11 +19,10 @@ import { useForm } from '../form';
 
 export interface DropdownQuestion {
   title: string;
+  description?: string;
   isRequired: boolean;
   key: QuestionKey;
   choices: Choice[];
-  description?: string;
-  prepopulateKey?: string;
```
MarkDKoslow: Should this have been removed?

no-config-components/form/index.tsx
```diff
@@ -109,6 +109,10 @@ const generateStyles: StyleGenerator = () => {
     formBodyStyles: css`
       ${responsiveWidthSnippet}
       margin: 16px auto 64px;
+
+      & > * {
```
MarkDKoslow: Does this mean these styles would get applied to all elements?

mellogarrett: Just the direct child elements, so the container divs for each question type component

no-config-components/dropdown-question/index.tsx
mellogarrett: Yea, we won't prepopulate the same way we do now. We'd instead fetch some data in the top-level form component and pass it into this component props. Something like:

```ts
const Form = () => {
  const choices = atlas.getChoices();

  return (
    <FormComponent title={'No Config' as any}>
      <CheckboxQuestionComponent
        question={{
          title: 'Checkbox',
          choices,
          isRequired: false,
          key: 'test-checkbox' as any,
        }}
      ></CheckboxQuestionComponent>
    </FormComponent>
  );
};
```

PR #430, opened 2020-02-12
CORE-1426 Simplify Peer Review Form

mellogarrett:
## Jira Issue

[CORE-1426]

## Description

Simplify Peer Review Form

## Type of change

[CORE-1426]: https://lambdaschool.atlassian.net/browse/CORE-1426


types/question-keys.ts
```diff
@@ -136,6 +136,7 @@ export const peerReviewKeys = [
   'otherInternal',
   'tags',
   'reviewEscalation',
+  'overallFeedback',
```
MarkDKoslow: Note to delete the removed keys here as well

mellogarrett: I thought about this. We haven't actually removed these columns from the database yet so I opted to keep them, just in case we get a request to add the questions back a month from now 😆

PR #445, opened 2020-02-24
CORE-1960 Add no-config radio question

MarkDKoslow:
## Jira Issue
[CORE-1960](https://lambdaschool.atlassian.net/browse/CORE-1960)

## Description
Config-less radio question

## Type of change


no-config-components/radio-question/index.test.tsx
```diff
@@ -0,0 +1,151 @@
...
+            description: 'Please let us know',
+            choices: [
+              { key: 'yes', display: 'Yes', value: 'Yes', checked: false },
+              { key: 'no', display: 'No', value: 'No', checked: false },
+            ],
+            isRequired: true,
+            key: 'test' as any,
+          }}
+        />
+      );
```
mellogarrett: ^ add a 
```ts
throw new Error('wrong error')
``` 
here

no-config-components/radio-question/index.tsx
```diff
@@ -0,0 +1,56 @@
...
+import { QuestionTitleComponent } from '../../components/question-title';
+import { useForm } from '../form';
+import { useManagedQuestion } from '../../utils/use-managed-question';
+import { getQuestionByKey } from '../../utils/get-question-by-key';
+import { updateMultiOptionAction } from '../../store/actions';
+
+export interface RadioQuestion {
+  title: string;
+  description?: string;
+  isRequired: boolean;
```
mellogarrett: Not a blocker, but I think we should make this optional eventually and just assume it's not required if they leave it off.

no-config-components/radio-question/index.tsx
```diff
@@ -0,0 +1,56 @@
...
+  key: QuestionKey;
+  prepopulateKey?: string;
```
mellogarrett: Not needed

no-config-components/radio-question/index.tsx
MarkDKoslow: I think I'd rather it be explicit but don't have a strong preference here.

mellogarrett: Yea, maybe even change this up to be an opt-in `isOptional` flag and default everything to required?

no-config-components/radio-question/index.test.tsx
```diff
@@ -0,0 +1,153 @@
...
+
+      throw new Error('Wrong Error');
+    } catch (err) {
+      expect(err.message).toBe('FormContext is undefined. You can only use this hook in a child of <FormComponent/>.');
+      expect(err.message).not.toBe('wrong error');
```
mellogarrett: Lines 30 & 33 should match

MarkDKoslow: Ah good catch

no-config-components/radio-question/index.tsx
MarkDKoslow: Yeah either one works for me. I'd vote we migrate all the components and then do a round of refactors/edits for shared properties

PR #455, opened 2020-03-02
Add no-config TextQuestion

mellogarrett:
## Jira Issue

n/a

## Description

- Add no-config `<TextQuestionComponent variant="short | long"/>`
- Fix bug in no-config `<RadioQuestionComponent/>`

## Type of change


no-config-components/text-question/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+        dispatch(updateErrorsAction(validationErrors, question.key));
+      }
+    }
+  };
+
+  return (
+    <div className={classname} {...rest}>
+      <QuestionTitleComponent title={title} description={description} isRequired={isRequired} />
+      {variant === 'long' ? (
+        <TextField
```
boblauer: Looks like there's a lot of overlap between these two, we can combine them into one:

```jsx
<TextField
  variant="outlined"
  multiline={variant === 'long'}
  rows={variant === 'long' ? 4 : 1}
  fullWidth
  type="text"
  value={value || ''}
  required={isRequired}
  onChange={onChange}
  onBlur={onBlur}
  error={showErrorMessage}
  helperText={showErrorMessage && firstError}
  inputProps={{ 'aria-label': title }}
/>
```

mellogarrett: Yea good call, looks like we can do this! We ran into a Typescript error a while back that caused us to split it out like this, but I can't reproduce it now so... 🤷‍♂

no-config-components/radio-question/index.test.tsx
```diff
@@ -102,7 +103,7 @@ describe('RadioQuestion', () => {
             title: 'Did you pair program today?',
             choices: [
               { key: 'yes', display: 'Yes', value: 'Yes', checked: false },
-              { key: 'no', display: 'No', value: 'No', checked: false },
+              { key: 'no', display: 'No', value: 'No', checked: true },
```
mellogarrett: This small update ensures we are changing the value, rather than just setting an initial value. This small distinction caught the incorrect action being dispatched.

mellogarrett:
Also changed validations to run `onBlur` instead of `onChange`. This is a better user experience IMO.

no-config-components/text-question/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+  const questionFromStore = getQuestionByKey(form, question.key) as TextQuestion;
+  if (!questionFromStore) return null;
+
+  const { title, isRequired, description, value, errors } = questionFromStore;
+
+  const classname = variant === 'long' ? 'long-text-question' : 'short-text-question';
+  const firstError = errors && errors[0];
+  const showErrorMessage = !!value && !!firstError;
+  const onChange = (e: ChangeEvent<HTMLInputElement | HTMLTextAreaElement>) => {
```
boblauer: I try to avoid using the event name as the function name, because it tells us _when_ the function will be called (which we can determine from L70 below) rather than _what_ the function is doing.

In this case, it looks like `onChange` is updating the form state, and `onBlur` is validating the input. I think `dispatchStateChange` and `validateInput` would be more appropriate names here.

no-config-components/text-question/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+interface TextQuestionProps {
+  question: TextQuestion;
+  variant: 'short' | 'long';
+  style?: CSSProperties;
+  className?: string;
+}
+
+export const TextQuestionComponent = ({ question, variant, ...rest }: TextQuestionProps) => {
+  const [form, dispatch] = useForm();
+  const longTextQuestion: LongTextQuestion = { ...question, type: QuestionType.LongText };
```
boblauer: We could consider consolidating all of our variant changes into a single place and use array destructuring to define multiple variables, rather than changing `variant === 'long'` throughout the code.

```ts
const [ textQuestion: LongTextQuestion | ShortTextQuestion, classname, isMultiLine, textFieldRows ] = variant === 'long' ? 
[ { ...question, type: QuestionType.LongText }, 'long-text-question', true, 4 ] :
[ { ...question, type: QuestionType.ShortText }, 'short-text-question', false, 1 ];
```

I've never actually used this pattern before, just something I thought up just now. Not sure if the consolidation is worth the indirectness. Could also destructure an object instead of an array, I don't know if that would be any better. 🤷‍♂

no-config-components/text-question/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+  const { title, isRequired, description, value, errors } = questionFromStore;
+
+  const classname = variant === 'long' ? 'long-text-question' : 'short-text-question';
+  const firstError = errors && errors[0];
+  const showErrorMessage = !!value && !!firstError;
+  const onChange = (e: ChangeEvent<HTMLInputElement | HTMLTextAreaElement>) => {
+    dispatch(updateFreeformAction(e.target.value, question.key));
+  };
+  const onBlur = (e: FocusEvent<HTMLInputElement | HTMLTextAreaElement>) => {
+    if ('validators' in question && question.validators) {
```
boblauer: I'm curious, why the `in` check here? If `question.validators` is true, `'validators' in question` will also be true.

mellogarrett: Yea, you're right.

no-config-components/text-question/index.tsx
mellogarrett: Yea I like this, but the type alias here doesn't work (`: LongTextQuestion | ShortTextQuestion`) and we have a lint rule against doing it on the right-hand side of the `=` with `as LongTextQuestion ` or `as ShortTextQuestion`

boblauer: Looks like you'd have to do:

```ts
const [ textQuestion, classname, isMultiLine, textFieldRows]: [ShortTextQuestion | LongTextQuestion] = [...]
```

PR #461, opened 2020-03-05
CORE-1959 Functioning No-config Forms

mellogarrett:
## Jira Issue

[CORE-1959]

## Description

Holy moly, where to begin... First off, I'm sorry 😅 
#### - Refactored `<FormComponent/>` so it now owns adding questions to the store (rather than the `useManagedQuestion` hook. 

This is actually a huge performance improvement because we don't have to go through a rendering cycle before we add the questions to the store. Before this, there was a weird flash in the UI for the second where no questions had been added to the store. 

This adding of questions to the store after a render cycle also caused a lot of issues with the submit button once I started to implement it. Because the form is _technically_ valid with no questions, the submit button would be enabled before the questions had been rendered/added to the store. To fix the issue, questions are now added in `<FormComponent/>` based on its `children`'s `question` prop. 

After making that change, there wasn't a great way to determine the `QuestionType` for each of the `<FormComponent/>` children 😓. I did some searching and there were only a couple of places that really depended on `QuestionType`
1) The code to merge in question options from the backend. This was easy to nix because we simply won't use that logic in no-config land. 
2) We used `QuestionType` to get a question's answer when a user clicked submit; For some question types it was derived from `choices`, for others it was `value`, etc. This caused the most pain and required me to refactor the store code. This probably caused the most file changes, but the outcome is much cleaner; There is now no such thing as an `AnswerGetter`. The answer is just always stored on the question as `answer`, regardless of the `QuestionType`. Simple enough 🤷‍♂. We also don't have to define `{ ..., checked: false }` on every multiple choice question's choice now 🎉 

#### - Forked the `store` code, several TypeScript interfaces, utils, etc.

There are now `NoConfig` versions of a lot of these things, in addition to the pre-existing versions of themselves.

#### - Update NextJS and TypeScript versions to support some cool new features, namely nullish coalescing (which VSCode kept adding to my code via autocomplete!?).

This new feature set will still live alongside our current implementation of Forms and _should_ have no affect on the way things currently operate. I will thoroughly test that theory in production.

## Type of change

[CORE-1959]: https://lambdaschool.atlassian.net/browse/CORE-1959


no-config-components/text-question/index.tsx
```diff
@@ -72,10 +53,12 @@ export const TextQuestionComponent = ({ question, variant, ...rest }: TextQuesti
...
         fullWidth
         type="text"
-        value={value || ''}
+        value={answer || ''}
         required={isRequired}
-        onChange={updateValue}
-        onBlur={runValidations}
+        onChange={(e) => {
+          updateValue(e);
+          runValidations(e);
```
MarkDKoslow: Did we have to move this from onBlue?

no-config-store/reducers/update-multi-option.ts
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  const indexOfNewValue = oldAnswer.indexOf(incomingChoiceValue);
+
+  if (indexOfNewValue === -1) {
+    return { ...question, answer: [...oldAnswer, incomingChoiceValue] };
+  }
+  const newQuestion = { ...question, answer: oldAnswer.filter((answer) => answer !== incomingChoiceValue) };
+  if (newQuestion.answer.length === 0) {
+    delete newQuestion.answer;
```
MarkDKoslow: woah never knew you could do this

no-config-components/text-question/index.tsx
boblauer: No, it was so that as the user types the field will show as valid or not. It fixed the use case where e.g. the submit button was enabled, the user is typing into the last field but types something invalid, and when clicking on the submit button the invalid field blurs and the button is disabled 1ms before they click the button.

MarkDKoslow: Oh wow yeah definitely would not have thought of that case. Makes sense 👍

PR #467, opened 2020-03-10
No-config conditionally visible questions

mellogarrett:
# [N/A] No-config conditionally visible questions

Type of change: [New feature]

## Description

Update `isVisible` prop on Question components to take a callback function with the signature:
```ts
(form: NoConfigForm) => boolean
```

## Launch Checklist

- [ ] Start Launch checklist
- [ ] Empty checkbox to remind me to do something


no-config-components/checkbox-question/index.tsx
```diff
@@ -31,7 +31,8 @@ export const CheckboxQuestionComponent = ({ question, ...rest }: CheckboxQuestio
   const questionFromStore = getNoConfigQuestionByKey(form, question.key) as NoConfigCheckboxQuestion;
   if (!questionFromStore) return null;
 
-  const { title, choices, description, isRequired, answer } = questionFromStore;
+  const { title, choices, description, isRequired, answer, isVisible } = questionFromStore;
+  if (isVisible && isVisible(form) === false) return <div data-testid="hidden-question" />;
```
boblauer: This feels off. Rendering the element only for testing purposes seems wrong.

Can we instead have `waitForElement` wait for the `FormComponent` to have been rendered, or to add some element to the bottom of the form that we key off of instead:

```js
    const { getByTestId, queryByText } = render(
      <FormComponent title={'form title' as any}>
        <DropdownQuestionComponent
          question={{
            title: 'test title' as any,
            description: 'Test description',
            isRequired: true,
            choices: [{ key: 'a', value: 'a' }],
            answer: 'a',
            key: 'test' as any,
            isVisible: () => false,
          }}
        />
        <div data-testid="form-ready" />
      </FormComponent>
    );
```

mellogarrett: I agree it does feel wrong and we _could_ do what you're proposing, though I don't see how it's much different than what I'm currently doing and potentially introduces bugs. 

For example, in an earlier implementation when we had the Question components responsible for adding themselves to the store, you'd be able to find `<div data-testid="form-ready" />` before the Question component had rendered. 

Testing for the absence of something in an async world is not a very safe test. I'd much rather add these elements and know without a doubt the correct logic had executed in my components.

boblauer: I don't think it would introduce a bug. As long as we have one test testing for the presence of the element, and one test testing for its absence, then e.g. if the questions rendered null on first pass, our tests checking for the presence of the elements would fail.

Do these tests need to `await`? It's not clear to me why we need `waitForElement` at all.

mellogarrett: RIght, it wouldn't introduce a bug with the current implementation. But it leaves gaps where bugs could creep in later. Testing for the absence of something is brittle IMO.

mellogarrett: I don't believe we need `waitForElement` anymore. We only needed that when questions would first render null, then once the store had been updated render the correct markup. Adding questions to the store in `<FormComponent/>` solved that problem. I'll update tomorrow.

boblauer: I'm having a hard time understanding why testing for the absence of something is brittle. It seems equally as brittle as testing for the presence of something. I don't see any difference between the two, and I'm trying to figure out what I'm missing.

mellogarrett: Because as long as you're looking for the absence of something, you never know when to stop looking. It may render as soon as the "looking" has timed out. 

When these components were rendered asynchronously I had to set some arbitrary time limit to look for the component within. By default, it was just below Jest's test timeout threshold (5s).

*Currently* this is not a problem, because we found a way to render everything synchronously. If the implementation changed again to be asynchronous, we'd be back in the same boat.

In this case I also find it nice to know *why* we're returning `null`, hence the `hidden-question` div instead.

mellogarrett: If you still want me to change it, I will no problem. As it stands today (everything synchronously rendered), testing for `null` would be perfectly valid.

boblauer: I don't hate it. I wish there was a better html element than div. I considered `<meta>`, but it looks like it's only valid within `<head>`

boblauer: Ideally we could render an element that has no effect on the DOM at all, but I don't think such an element exists for this purpose.

PR #474, opened 2020-03-16
CORE-2040 Add no-config QuestionGroup

mellogarrett:
## Description

- Add no-config `<QuestionGroup/>` component
- Add no-config Sprint Review form as a starting point

## Type of Change 

- New feature (non-breaking change which adds functionality)


pages/no-config-sprint-review.tsx
```diff
@@ -0,0 +1,206 @@
...
+          
+    Student can access this feedback through their dashboard.`,
+          isRequired: true,
+          key: 'studentVisibleSkillsFeedback',
+        }}
+      />
+      <QuestionGroup
+        title="Unit Review"
+        isVisible={(form) => {
+          const parentQuestion = getNoConfigQuestionByKey(form, 'sprintSubmissionId') as NoConfigDropdownQuestion;
```
boblauer: I see this pattern of having to get the question by key coming up in pretty much every `isVisible` and other component functions that are passed the entire form.

We might want to consider adding a method directly to the `form` object, something like `form.getQuestion('sprintSubmissionId')`.

pages/no-config-sprint-review.tsx
```diff
@@ -0,0 +1,206 @@
...
+          const selectedChoice = parentQuestion.choices.find((choice) => choice.value === parentQuestion.answer);
```
boblauer: Getting the selected answer to a question with `choices` is going to happen often as well, might want to add a helper function for that as well. Or perhaps we could add something like `parentQuestion.selectedChoice.metadata`?

pages/no-config-sprint-review.tsx
boblauer: `parentQuestion` feels vague. Perhaps `sprintSubmission`?

no-config-components/form/index.tsx
```diff
@@ -34,15 +60,28 @@ export const useForm = () => {
...
 
+function getQuestionsFromChildren(children: ReactNode): NoConfigQuestion[] {
+  let questions: NoConfigQuestion[] = [];
+  let questionComponentChildren: GenericQuestionComponent[] = [];
+
+  Children.forEach(children, (child) => {
+    if (isQuestionComponent(child)) {
+      questionComponentChildren.push(child);
+    } else if (hasChildren(child)) {
+      questions = [...questions, ...getQuestionsFromChildren(children)]; // recursively add questions for entire component tree
```
MarkDKoslow: ⚒ Would this be an infinite loop? I think we'd just want to pass in the children for `child`

mellogarrett: Good catch.

no-config-components/question-group/index.tsx
```diff
@@ -0,0 +1,39 @@
...
+  children?: ReactNode;
+  style?: CSSProperties;
+  className?: string;
+}
+
+export const QuestionGroup = ({ title, isVisible, children, style, className }: NoConfigQuestionGroupProps) => {
+  const [form] = useForm();
+  const { questionGroupTitleStyles } = generateStyles();
+
+  if (isVisible && isVisible(form) === false) return <div data-testid="hidden-question-group" />;
```
MarkDKoslow: This may be something to come back to later, but wouldn't we need to pass in more information into isVisible? Like the current questionGroup title/key?

mellogarrett: No, since it's just a callback. You decide how to use it in the form page code.

pages/no-config-sprint-review.tsx
mellogarrett: It does get a bit hairy here because it could be plural `selectedChoices` for checkbox or multi-select dropdowns.

pages/no-config-sprint-review.tsx
mellogarrett: done

utils/get-no-config-question-by-key/index.ts
```diff
@@ -1,15 +1,7 @@
...
-  const flattenedQuestions = form.questions.reduce(
-    (questions: NoConfigQuestion[], questionOrQuestionGroup) =>
-      isNoConfigQuestionGroup(questionOrQuestionGroup)
-        ? [...questions, ...questionOrQuestionGroup.questions]
-        : [...questions, questionOrQuestionGroup],
-    []
-  );
-
-  return flattenedQuestions.find((question) => question.key === key);
+  return form.questions.find((question) => question.key === key);
```
boblauer: Is this file even needed anymore, or can we always use `form.getQuestionByKey` everywhere we're currently using this function?

pages/no-config-sprint-review.tsx
mellogarrett: I think what would work well is if we store an `.answerMetadata` on the question and update it when we update `.answer` in the reducers. I'm envisioning just a flat object that would aggregate any metadata from selected choice(s). 
```ts
type NoConfigQuestion {
  ...
  answerMetadata: {
    [key: string]: string | number | boolean
  }
}
```
Then we'd have a uniform way of checking for values in metadata.

Though, that is probably out of scope for this PR.

utils/get-no-config-question-by-key/index.ts
mellogarrett: Yea, can probably remove altogether.

mellogarrett: done

pages/no-config-sprint-review.tsx
mellogarrett:
```tsx
<Question
  question={{
    ...
    title: 'Some Unit Question',
    isVisible: form => 
      form.getQuestion('sprintSubmission').answerMetadata['hasUnitSubmission'] === true
  }}
/>
```

boblauer:
Do you know what the failing tests are about?

mellogarrett:
Not right off, no.

mellogarrett:
Fixed the tests. Added a `createForm` function that just adds the `getQuestion` for you to be used in the `<FormComponent/>` and tests alike.

PR #511, opened 2020-04-16
Bug fix: Enable choosing a Dropdown option with the keyboard

mellogarrett:
## Description

Enable choosing a Dropdown option with the keyboard only

## Type of Change 

- Bug fix (non-breaking change which fixes an issue)
- New feature (non-breaking change which adds functionality)


no-config-components/dropdown-question/index.tsx
```diff
@@ -109,9 +115,7 @@ export const DropdownQuestionComponent = ({ question, styles, className }: Dropd
                 key={`chip-${choice.value}${choice.key}`}
                 label={choice.display || choice.value}
                 onDelete={() => {
-                  if (choice) {
-                    dispatch(updateMultiOptionAction(choice.value.toString(), questionKey));
-                  }
```
MarkDKoslow, line 114: Why this change?

mellogarrett: Choice is always defined here

MarkDKoslow: Oh I see, thought this was deleting from the search text not the chip

